PR #1, opened 2019-05-09
Add secp256k1-zkp and schnorrsig modules

jonasnick:
~~WIP because this requires a new rust-secp256k1 release (with https://github.com/rust-bitcoin/rust-secp256k1/pull/105 and https://github.com/rust-bitcoin/rust-secp256k1/pull/101).~~

This PR adds 3 crates, a no_std and low-level `zkp-sys`, higher level `zkp` and a crate `-dev` for dev dependencies.


jonasnick:
un-WIP'ed because rust-secp256k1 0.13 is released

secp256k1-zkp-dev/src/lib.rs
```diff
@@ -0,0 +1,58 @@
...
+// any warranty.
+//
+// You should have received a copy of the CC0 Public Domain Dedication
+// along with this software.
+// If not, see <http://creativecommons.org/publicdomain/zero/1.0/>.
+//
+
+/// The functions in this module are copied from secp256k1 because they can only be used when
+/// compiled with the `rand` feature. But the other libraries need them as a dev-dependency for
+/// `cargo test` also when `rand` is not enabled.
```
apoelstra: Is this https://github.com/rust-lang/cargo/issues/1796 ? Can you link to that issue in this comment?

secp256k1-zkp-dev/src/lib.rs
```diff
@@ -0,0 +1,58 @@
...
+use secp256k1::{PublicKey, Secp256k1, SecretKey, Signing};
+
+fn random_32_bytes<R: Rng>(rng: &mut R) -> [u8; 32] {
+    let mut ret = [0u8; 32];
+    rng.fill_bytes(&mut ret);
+    ret
+}
+
+trait New {
+    fn new<R: Rng>(rng: &mut R) -> SecretKey;
```
apoelstra: I think you want `impl Distribution<SecretKey> for Standard` rather than this trait.

See the second example in https://docs.rs/rand/0.6.5/rand/distributions/index.html

secp256k1-zkp-dev/src/lib.rs
```diff
@@ -0,0 +1,58 @@
...
+
+impl New for SecretKey {
+    /// Creates a new random secret key.
+    #[inline]
+    fn new<R: Rng>(rng: &mut R) -> SecretKey {
+        let mut key = SecretKey::from_slice(&random_32_bytes(rng));
+        while key.is_err() {
+            key = SecretKey::from_slice(&random_32_bytes(rng));
+        }
+        key.unwrap()
```
apoelstra: It avoids an uwrap to do
```
loop {
    if let Ok(key) = SecretKey::from_slice(&random_32_bytes(rng)) {
        return key;
    }
}
```
which is a little more rustic.

secp256k1-zkp-sys/build.rs
```diff
@@ -0,0 +1,83 @@
...
+        .include("depend/secp256k1-zkp/include")
+        .include("depend/secp256k1-zkp/src")
+        .debug(true)
+        .flag_if_supported("-Wno-unused-function") // some ecmult stuff is defined but not used upstream
+        .define("SECP256K1_BUILD", Some("1"))
+        // TODO these three should be changed to use libgmp, at least until secp PR 290 is merged
+        .define("USE_NUM_NONE", Some("1"))
+        .define("USE_FIELD_INV_BUILTIN", Some("1"))
+        .define("USE_SCALAR_INV_BUILTIN", Some("1"))
+        .define("USE_ENDOMORPHISM", Some("1"))
```
apoelstra: We should drop this for now because gmax is concerned about patent aggression related to it.

secp256k1-zkp-sys/src/lib.rs
```diff
@@ -0,0 +1,93 @@
...
+
+#[macro_use]
+mod macros;
+pub mod ffi;
+pub mod schnorrsig;
+mod types;
+
+/// The purpose of this rewriteable structure is to replace dynamic memory allocations in the
+/// libsecp256k1-zkp library. See its documentation for further information.
+pub struct ScratchSpace {
```
apoelstra: I think we can impl both `Send` (no external references) and `Sync` (no interior mutability) on this.

secp256k1-zkp-sys/src/schnorrsig/mod.rs
```diff
@@ -0,0 +1,334 @@
...
+
+use core::{fmt, ptr};
+
+use secp256k1::key::{PublicKey, SecretKey};
+use secp256k1::{self, Message, Secp256k1, Signing, Verification};
+
+use super::ScratchSpace;
+
+/// A Schnorrsig error. This does not implement `std:error::Error` because it's not available with
+/// `no_std`.
```
apoelstra: Can you add an implementation of `error::Error` gated on the std feature?

apoelstra: Oh, I see, it is implemented on the type in the "real" library.

secp256k1-zkp-sys/build.rs
jonasnick: I just copied it from rust-secp. Is it a concern there also?

secp256k1-zkp-dev/src/lib.rs
jonasnick: Ah, yes.

secp256k1-zkp-dev/src/lib.rs
jonasnick: Ok (this was all just copied from rust-secp)

secp256k1-zkp-dev/src/lib.rs
jonasnick: I'm getting ` E0117` ("impl doesn't use types inside crate") because `SecretKey` isn't defined in this crate. This looks ergonomic but a downside is that copy & pasted code starts diverging from rust-secp.

secp256k1-zkp-sys/src/lib.rs
jonasnick: Perhaps I shouldn't do that because I don't understand the implications. Afaics `Send` is fine but `Sync` isn't (`Sync [...] answers the question: “if multiple threads were all trying to access this data, would it be safe?”` from https://manishearth.github.io/blog/2015/05/30/how-rust-achieves-thread-safety/).

secp256k1-zkp-dev/src/lib.rs
apoelstra: Can you PR to rust-secp? Or then do we run into the problem with dev-dependencies features?

secp256k1-zkp-sys/build.rs
apoelstra: Yes, I thought we'd removed it from rust-secp.

secp256k1-zkp-sys/src/lib.rs
apoelstra: Yes, if multiple threads are trying to access the data it's fine. They can't mutate it because it's impossible for multiple threads to have ownership or `&mut` access simultaneously.

secp256k1-zkp-dev/src/lib.rs
jonasnick: Oh, yes we'd run into that problem.

secp256k1-zkp-dev/src/lib.rs
jonasnick: fixed

secp256k1-zkp-sys/src/lib.rs
jonasnick: done

apoelstra:
ACK

jonasnick:
squashed

PR #9, opened 2021-01-06
Prepare for new releases after replacing master

jonasnick:


thomaseizinger (changes requested):
Thanks for updating the metadata.

I think we should also re-run the vendor script to make sure the version number in the symbols aligns with the version of the crate.

jonasnick:
Okay, done.

thomaseizinger (approved):
tACK 1148ea4

I ran the vendor script locally and it produces the same result.

PR #10, opened 2021-04-01
Fix Pedersen Commitment deserialization

luckysori:
The first commit identifies the bug and the second fixes it. If this PR gets approved I'll squash them together.

### DIFF
--- a/src/zkp/pedersen.rs
+++ b/src/zkp/pedersen.rs
@@ -188,9 +188,9 @@ impl fmt::Display for PedersenCommitment {
 impl str::FromStr for PedersenCommitment {
     type Err = Error;
     fn from_str(s: &str) -> Result<Self, Error> {
-        let mut res = [0; 64];
+        let mut res = [0; 33];
         match from_hex(s, &mut res) {
-            Ok(64) => Self::from_slice(&res[0..64]),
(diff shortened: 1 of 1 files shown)


jonasnick (commented):
Looks good to me. Not sure what's wrong with CI.

thomaseizinger (approved):
Interesting, what a stupid copy-paste bug :D

I have two concerns:

1. If you want to test the serialization, can we test it using `serde_test` please AND make sure we cover the human-readable / non-human-readable part as well?
2. Given that the bug occurred within the `str::FromStr` implementation, I would actually prefer if we test that instead of the `serde` impl which just happened to forward to that. That would also be good because it ensures that we are actually upholding the invariant of `Display`/`FromStr` being round-trippable.

Happy to merge this if it is urgent, as long as there is a follow-up!

thomaseizinger (approved):

thomaseizinger:
Thanks for updating the tests!

thomaseizinger:
@jonasnick I merged this based on the assumption that https://github.com/ElementsProject/rust-secp256k1-zkp/pull/10#pullrequestreview-626109161 is an ACK :)

jonasnick:
@thomaseizinger yup all good, my comment was more like "one real ACK is fine for this"

PR #11, opened 2021-04-07
Add ecdsa adaptor

Tibo-lg:
As ECDSA adaptor signatures PR has been merged upstream (https://github.com/ElementsProject/secp256k1-zkp/pull/117) I tried to port the code I had for them in a fork of rust-secp256k1 here (hoping that it could be useful to others).


src/zkp/ecdsa_adaptor.rs
```diff
@@ -0,0 +1,206 @@
...
+        let adaptor_sig = EcdsaAdaptorSignature::adaptor_encrypt(&secp, &msg, &seckey, &adaptor);
+
+        assert!(adaptor_sig
+            .adaptor_verify(&secp, &msg, &pubkey, &adaptor)
+            .is_ok());
+        assert!(adaptor_sig
+            .adaptor_verify(&secp, &msg, &adaptor, &pubkey)
+            .is_err());
+        let sig = adaptor_sig.adaptor_decrypt(&adaptor_secret);
+        #[cfg(not(rust_secp_fuzz))]
```
Tibo-lg, line 199: This doesn't feel super clean, but I don't have a very good idea on how to implement a version of `adaptor_decrypt` that would be compatible with the fuzz version of `verify` since it doesn't have access to the `msg`.

thomaseizinger: I'd rather have us disable to whole test to be honest. I am not sure we need to run them as part of fuzzing.

src/zkp/ecdsa_adaptor.rs
```diff
@@ -0,0 +1,206 @@
...
+                pubkey.as_c_ptr(),
+                msg.as_c_ptr(),
+                encryption_key.as_c_ptr(),
+            );
+            return if res == 1 {
+                Ok(())
+            } else {
+                Err(Error::Upstream(UpstreamError::IncorrectSignature))
+            };
+        }
```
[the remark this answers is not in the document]
thomaseizinger: Same goes for the other functions as well.

src/zkp/ecdsa_adaptor.rs
```diff
@@ -0,0 +1,206 @@
...
+            assert_eq!(
+                1,
+                ffi::secp256k1_ecdsa_adaptor_recover(
+                    *secp.ctx(),
+                    data.as_mut_c_ptr(),
+                    sig.as_c_ptr(),
+                    self.as_c_ptr(),
+                    encryption_key.as_c_ptr()
+                )
+            );
```
thomaseizinger, line 146: I haven't dug into the details of the C implementation. What happens if we pass a bad signature here?

My assumption would be that the function returns 0 in which case this would panic. I don't think we should panic in this case but rather return an error along the lines of `FailedToRecover` because the signature doesn't fit to the adaptor.

src/zkp/ecdsa_adaptor.rs
```diff
@@ -0,0 +1,206 @@
...
+            let mut signature = ffi::Signature::new();
+            assert_eq!(
+                1,
+                ffi::secp256k1_ecdsa_adaptor_decrypt(
+                    ffi::secp256k1_context_no_precomp,
+                    &mut signature,
+                    decryption_key.as_c_ptr(),
+                    self.as_c_ptr(),
+                )
+            );
```
thomaseizinger, line 122: Similar to `recover`: Will this return 0 and hence panic if we pass a non-fitting secret key?

thomaseizinger (changes requested):
Thanks for the contribution @tibo-lg!
Definitely most welcome :)

I've left some comments regarding naming, `unsafe`-block size and very importantly: panic behaviour.

src/zkp/ecdsa_adaptor.rs
```diff
@@ -0,0 +1,206 @@
...
+                *secp.ctx(),
+                self.as_c_ptr(),
+                pubkey.as_c_ptr(),
+                msg.as_c_ptr(),
+                encryption_key.as_c_ptr(),
+            );
+            return if res == 1 {
+                Ok(())
+            } else {
+                Err(Error::Upstream(UpstreamError::IncorrectSignature))
```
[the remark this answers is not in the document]
thomaseizinger: If we introduce another variant for handling the cases of passing in a bad secret-key or signature, we might also want to use that because strictly speaking the adaptor sig is not "malformed" it just doesn't fit together.

Open for suggestions here in how to best handle this. I do think that `Error::Upstream` is not a good choice here because that variant only exists so that we can call into upstream functionality and properly return those errors. We shouldn't be constructing those ourselves because it is misleading as they don't actually originate upstream then.

src/zkp/ecdsa_adaptor.rs
Tibo-lg: Done, sounds good to me.

src/zkp/ecdsa_adaptor.rs
```diff
@@ -0,0 +1,206 @@
...
+use {Message, Signing};
+use {Signature, Verification};
+
+/// Represents an adaptor signature and dleq proof.
+#[derive(Debug, PartialEq, Clone, Copy, Eq)]
+pub struct EcdsaAdaptorSignature(ffi::EcdsaAdaptorSignature);
+
+impl fmt::LowerHex for EcdsaAdaptorSignature {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        for ch in self.0.to_bytes().iter() {
```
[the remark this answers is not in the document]
Tibo-lg: Done, indeed it's better not to copy

src/zkp/ecdsa_adaptor.rs
```diff
@@ -0,0 +1,206 @@
...
+    /// Obtains a raw mutable pointer suitable for use with FFI functions
+    #[inline]
+    pub fn as_mut_ptr(&mut self) -> *mut ffi::EcdsaAdaptorSignature {
+        &mut self.0
+    }
+}
+
+impl EcdsaAdaptorSignature {
+    /// Creates an adaptor signature along with a proof to verify the adaptor signature.
+    pub fn adaptor_encrypt<C: Signing>(
```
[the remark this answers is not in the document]
Tibo-lg: Good point, updated names

src/zkp/ecdsa_adaptor.rs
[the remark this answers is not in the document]
Tibo-lg: Done

src/zkp/ecdsa_adaptor.rs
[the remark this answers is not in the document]
Tibo-lg: I see your point, thanks for the explanation. I changed to use `InvalidEcdsaAdaptorSignature` for all errors related to this module for now. I'm not sure it's the best, but at the same time I'm also not sure it would be better to have `FailedToRecover`, `FailedToDecrypt`... especially thinking that we can't really tell what the issue is apart that the native code returned an error. If you prefer the multiple error approach I don't mind updating it though.

src/zkp/ecdsa_adaptor.rs
Tibo-lg: Your assumption is correct, thanks for catching this. I updated to return `Result<>` from all functions. See previous comment regarding error naming.

src/zkp/ecdsa_adaptor.rs
Tibo-lg: updated

src/zkp/ecdsa_adaptor.rs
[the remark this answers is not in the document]
Tibo-lg: Though note that for `decrypt` all the code is kept inside the `unsafe` because `ffi::Signature::new()` is also unsafe. It's possible to do something like:
```rust
        let (ret, signature) = unsafe {
            let mut signature = ffi::Signature::new();
            let ret = ffi::secp256k1_ecdsa_adaptor_decrypt(
                ffi::secp256k1_context_no_precomp,
                &mut signature,
                decryption_key.as_c_ptr(),
                self.as_c_ptr(),
            );
            (ret, signature)
        };

        if ret != 1 {
            return Err(Error::InvalidEcdsaAdaptorSignature);
        }

        Ok(Signature::from(signature))
``` 
but I felt this is a bit hard to read. Let me know if you prefer that or have a better idea.

src/zkp/ecdsa_adaptor.rs
```diff
@@ -0,0 +1,317 @@
...
+                enckey.as_c_ptr(),
+                msg.as_c_ptr(),
+                None,
+                ptr::null_mut(),
+            )
+        };
+
+        if ret != 1 {
+            return Err(Error::InvalidEcdsaAdaptorSignature);
+        }
```
thomaseizinger, line 106: Unless I am mistaken and from looking at the C code, this function only ever fails if any of the arguments are null: https://github.com/ElementsProject/secp256k1-zkp/blob/f3708a1ecb445b1b05a0f8fcd1da6a88f83d89c4/src/modules/ecdsa_adaptor/main_impl.h#L149-L224

We statically guarantee that this is the case using the type system so I think we can `debug_assert!` this return value here instead of returning an error.

(Note to self: add fuzzing to this project so can make these kind of assumptions :grin: )

src/zkp/ecdsa_adaptor.rs
```diff
@@ -0,0 +1,317 @@
...
+    #[inline]
+    pub fn from_slice(data: &[u8]) -> Result<EcdsaAdaptorSignature, Error> {
+        match data.len() {
+            ECDSA_ADAPTOR_SIGNATURE_LENGTH => {
+                let mut ret = [0; ECDSA_ADAPTOR_SIGNATURE_LENGTH];
+                ret[..].copy_from_slice(data);
+                Ok(EcdsaAdaptorSignature(ffi::EcdsaAdaptorSignature::from(ret)))
+            }
+            _ => Err(Error::InvalidEcdsaAdaptorSignature),
+        }
```
thomaseizinger, line 66: Instead of just copying the bytes over, we could call [`secp256k1_ecdsa_adaptor_sig_deserialize`](https://github.com/ElementsProject/secp256k1-zkp/blob/f3708a1ecb445b1b05a0f8fcd1da6a88f83d89c4/src/modules/ecdsa_adaptor/main_impl.h#L30-L66) here and see if it parses successfully. Thoughts?

src/lib.rs
```diff
@@ -95,6 +95,8 @@ pub enum Error {
     InvalidRangeProof,
     /// Bad generator
     InvalidGenerator,
+    /// Bad AdaptorSignature
```
thomaseizinger, line 98:
```suggestion
    /// Given bytes don't represent a valid adaptor signature
```

src/zkp/ecdsa_adaptor.rs
```diff
@@ -0,0 +1,317 @@
...
+            let mut signature = ffi::Signature::new();
+            let ret = ffi::secp256k1_ecdsa_adaptor_decrypt(
+                ffi::secp256k1_context_no_precomp,
+                &mut signature,
+                decryption_key.as_c_ptr(),
+                self.as_c_ptr(),
+            );
+
+            if ret != 1 {
+                return Err(Error::InvalidEcdsaAdaptorSignature);
```
thomaseizinger, line 123: Looking at our other error variants we are actually quite detailed there. For example we have `CannotProveSurjection` and `InvalidSurjectionProof`.

I think we should do the same thing here and have variants for:

- `CannotDecryptAdaptorSignature`
- `CannotRecoverAdaptorSecret`
- `CannotVerifyAdaptorSignature` (this is different from invalid I'd say because one says that we could parse it and the other one says that the combination of signature, adaptor signature and keys doesn't work)

src/zkp/ecdsa_adaptor.rs
```diff
@@ -0,0 +1,317 @@
...
+        self.as_ptr()
+    }
+
+    fn as_mut_c_ptr(&mut self) -> *mut Self::Target {
+        self.as_mut_ptr()
+    }
+}
+
+impl EcdsaAdaptorSignature {
+    /// Creates an AdaptorSignature directly from a slice
```
thomaseizinger, line 56: The spelling of "AdaptorSignature" here is inconsistent with the rest of the file.

thomaseizinger (commented):
Thanks for working in the comments and adding all these tests!

I've left some more feedback. Everything related to the tests is more of a nit though !

@jonasnick Can you also have a look at this please?

thomaseizinger:
Confirmed that 7637c2e is the result of vendoring https://github.com/ElementsProject/secp256k1-zkp/commit/f3708a1ecb445b1b05a0f8fcd1da6a88f83d89c4.

src/zkp/ecdsa_adaptor.rs
Tibo-lg: That would be much better indeed but unfortunately AFAICT this function is not part of the public interface.

src/zkp/ecdsa_adaptor.rs
```diff
@@ -0,0 +1,317 @@
...
+    fn test_ecdsa_adaptor_signature() {
+        let secp = Secp256k1::new();
+
+        let (seckey, pubkey) = secp.generate_keypair(&mut thread_rng());
+        let (adaptor_secret, adaptor) = secp.generate_keypair(&mut thread_rng());
+        let msg = Message::from_slice(&[2u8; 32]).unwrap();
+        let adaptor_sig_res = EcdsaAdaptorSignature::encrypt(&secp, &msg, &seckey, &adaptor);
+        assert!(adaptor_sig_res.is_ok());
+
+        let adaptor_sig = adaptor_sig_res.unwrap();
```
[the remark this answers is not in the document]
Tibo-lg: Much better indeed thanks! I made the change everywhere.

src/zkp/ecdsa_adaptor.rs
```diff
@@ -0,0 +1,317 @@
...
+
+        assert!(adaptor_sig.verify(&secp, &msg, &pubkey, &adaptor).is_ok());
+        assert!(adaptor_sig.verify(&secp, &msg, &adaptor, &pubkey).is_err());
```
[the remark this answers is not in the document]
Tibo-lg: done

src/zkp/ecdsa_adaptor.rs
```diff
@@ -0,0 +1,317 @@
...
+        from_hex(
+            "8131e6f4b45754f2c90bd06688ceeabc0c45055460729928b4eecf11026a9e2d",
+            &mut msg_buf,
+        )
+        .unwrap();
+        let msg = Message::from_slice(&msg_buf).unwrap();
+        let pubkey = PublicKey::from_str(
+            "035be5e9478209674a96e60f1f037f6176540fd001fa1d64694770c56a7709c42c",
+        )
+        .unwrap();
```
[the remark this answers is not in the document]
Tibo-lg: It unfortunately still doesn't format on one line but still much easier to read I think, thanks I didn't know about this.

src/zkp/ecdsa_adaptor.rs
```diff
@@ -0,0 +1,317 @@
...
+        let adaptor_sig = EcdsaAdaptorSignature::from_str("03424d14a5471c048ab87b3b83f6085d125d5864249ae4297a57c84e74710bb6730223f325042fce535d040fee52ec13231bf709ccd84233c6944b90317e62528b2527dff9d659a96db4c99f9750168308633c1867b70f3a18fb0f4539a1aecedcd1fc0148fc22f36b6303083ece3f872b18e35d368b3958efe5fb081f7716736ccb598d269aa3084d57e1855e1ea9a45efc10463bbf32ae378029f5763ceb40173f").unwrap();
+
+        assert!(adaptor_sig
+            .verify(&secp, &msg, &pubkey, &encryption_key)
+            .is_ok());
+
+        let mut sig_buf = [0u8; 64];
+        from_hex("424d14a5471c048ab87b3b83f6085d125d5864249ae4297a57c84e74710bb67329e80e0ee60e57af3e625bbae1672b1ecaa58effe613426b024fa1621d903394", &mut sig_buf).unwrap();
```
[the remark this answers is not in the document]
Tibo-lg: I tried to use hex_literal but the latest version didn't work with 1.29 and I was lazy to find out which version would so I just ended up creating two small helper functions at the bottom of the tests. IMHO it's even cleaner and less code but let me know if you'd rather use hex_litteral.

src/zkp/ecdsa_adaptor.rs
```diff
@@ -0,0 +1,317 @@
...
+        assert!(secp.verify(&msg, &sig, &pubkey).is_ok());
+        let recovered_res = adaptor_sig.recover(&secp, &sig, &adaptor);
+        assert!(recovered_res.is_ok());
+        let recovered = recovered_res.unwrap();
+        assert_eq!(adaptor_secret, recovered);
+    }
+
+    #[test]
+    fn test_ecdsa_adaptor_signature_plain_valid() {
+        let secp = Secp256k1::new();
```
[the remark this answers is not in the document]
Tibo-lg: Ah thanks I hadn't properly understood this global-context thing, updated.

src/zkp/ecdsa_adaptor.rs
Tibo-lg: You're right, replaced as suggested.

src/lib.rs
Tibo-lg: Updated

src/zkp/ecdsa_adaptor.rs
Tibo-lg: Fair enough, I added the suggested errors. Maybe check the documentation for them because I'm not 100% confident it's good.

src/zkp/ecdsa_adaptor.rs
Tibo-lg: Thanks fixed (I assumed you meant it should be `EcdsaAdadptorSignature`?)

src/zkp/ecdsa_adaptor.rs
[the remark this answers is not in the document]
thomaseizinger: Fair enough. Personally, I don't think we need to stick to the MSRV for running the tests. Local helper functions are also fine though, the goal was to improve clarity :)

src/zkp/ecdsa_adaptor.rs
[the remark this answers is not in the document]
thomaseizinger: Has the added benefit that you don't need to import the `FromStr` trait!

src/zkp/ecdsa_adaptor.rs
thomaseizinger: You are right. It is not part of the [header](https://github.com/ElementsProject/secp256k1-zkp/blob/f3708a1ecb445b1b05a0f8fcd1da6a88f83d89c4/include/secp256k1_ecdsa_adaptor.h).

Let's leave it as is then!

src/zkp/ecdsa_adaptor.rs
thomaseizinger: Sorry I should have been clearer!

In the rest of the documentation, it is usually spelt "adaptor signature" and only here we use a PascalCase for some reason. Given that we are not referring to any type, I think the normal English spelling is preferable.

src/lib.rs
```diff
@@ -95,6 +95,14 @@ pub enum Error {
     InvalidRangeProof,
     /// Bad generator
     InvalidGenerator,
+    /// Given bytes don't represent a valid adaptor signature
+    InvalidEcdsaAdaptorSignature,
+    /// Failed to decrypt an adaptor signature because of an internal error within `libsecp256k1-zkp`
+    CannotDecryptAdaptorSignature,
+    /// Failed to recover an adaptor secret from an adaptor signature because of an internal error within `libsecp256k1-zkp`
+    CannotRecoverAdaptorSecret,
+    /// Given adaptor signature is not valid fo the provided combination of public key, encryption key and message
```
thomaseizinger, line 104:
```suggestion
    /// Given adaptor signature is not valid for the provided combination of public key, encryption key and message
```

thomaseizinger (commented):
Thank you!

Just one more typo in the docs!
Otherwise happy to merge this.

I'll give @jonasnick a couple of days in case he wants to have a look as well!

src/zkp/ecdsa_adaptor.rs
thomaseizinger: Documentation is good, thank you!

src/zkp/ecdsa_adaptor.rs
thomaseizinger: If you wanted to explicitly reference the type, you can use the square-brackets + backtick notation:

```rust
/// Creates an [`EcdsaAdaptorSignature`] directly from a slice
```

That will render to an actual link in rustdoc and on a recent toolchain, it will emit a warning if the type you are referencing doesn't actually exist!

Tibo-lg: Thanks that's neat, I changed to using that. I think it makes sense to reference the type here because it's somehow different than the `encrypt` function in the sense that the goal is to create the struct directly and not generate it. I don't have a strong opinion though so I can change it to plain English if you prefer.

src/lib.rs
Tibo-lg: Thanks for catching that, fixed!

thomaseizinger (approved):
utACK 1830ec6a59e502bbd3215c3a4ff71d560d6767b9

src/lib.rs
```diff
@@ -107,6 +115,10 @@ impl fmt::Display for Error {
             Error::CannotMakeRangeProof => "failed to generate range proof",
             Error::InvalidRangeProof => "failed to verify range proof",
             Error::InvalidGenerator => "malformed generator",
+            Error::InvalidEcdsaAdaptorSignature => "malformed ecdsa adaptor signature",
+            Error::CannotDecryptAdaptorSignature => "failed to decrypt adaptor signature",
+            Error::CannotRecoverAdaptorSecret => "failed to recover adaptor secret",
+            Error::CannotVerifyAdaptorSignature => "failed to verify adadptor signature",
```
jonasnick, line 121: nit: I can imagine this typo ("adadptor") making somewhen crazy when trying to grep error logs

src/zkp/ecdsa_adaptor.rs
```diff
@@ -0,0 +1,308 @@
+//! # ECDSA Adaptor
```
jonasnick, line 1: Unfortunately I think we need to mention the warning in secp here. Perhaps something like

```
WARNING: ECDSA adaptor signatures are insecure when the secret key is reused
in certain other crypto schemes. See
https://github.com/ElementsProject/secp256k1-zkp/blob/6955af5ca8930aa674e5fdbc4343e722b25e0ca8/include/secp256k1_ecdsa_adaptor.h#L14
for details.
```

jonasnick (commented):
Very nice!

ACK mod nits

src/lib.rs
Tibo-lg: Thanks fixed!

src/zkp/ecdsa_adaptor.rs
Tibo-lg: Good point added.

Tibo-lg:
Rebased on a7409c9bcfe121c1c0ad5bc7661a6b8cf8da5b3c

thomaseizinger:
tACK 309bbb4

Verified locally that f9b8dc7 just vendors https://github.com/ElementsProject/secp256k1-zkp/commit/f3708a1ecb445b1b05a0f8fcd1da6a88f83d89c4 and does nothing else.

jonasnick:
ACK 309bbb4f452e7b129d08c060d06054ec0c5740d5

PR #13, opened 2021-04-19
Prepare for minor release 0.3.0

jonasnick:


thomaseizinger (approved):
I guess even if there aren't any obvious breaking changes, it is better to err on the safe side of bumping the minor version instead of patch.

jonasnick:
As @sanket1729 pointed out to me, for some (rather esoteric afaics) matching on the Error type this can be a breaking change.

thomaseizinger:
> As @sanket1729 pointed out to me, for some (rather esoteric afaics) matching on the Error type this can be a breaking change.

Not very esoteric actually. An exhaustive `match` (ie not using `_ => {}`)  will fail to compile once a variant is added.

jonasnick:
Published on crates.io. @Tibo-lg fyi in case you want to use the crates.io release for ECDSA adaptor sigs.

Tibo-lg:
Thanks for the notice I do want to use it :)

PR #14, opened 2021-04-21
Add support for creating unblinded generators/commitments

sanket1729:
It is currently not possible to create generators and commitments with zero blinding factors because they require a `SecretKey` type that cannot be zero. 
Maybe the long-term solution is not to use the secret key type at all, and take input from a 32-byte array. But this should suffice for elements use-case where we need to create Pedersen commitments for explicit amounts with zero blinding factors.


apoelstra (approved):
ack e580d30ce7c14170fd5eeb3070195e8ef3576290

PR #16, opened 2021-04-26
Replace `SecretKey` based tweak methods with a new `Tweak type`

sanket1729:
While implementing additional validation in rust-elements, I realized that we need have zero value tweaks/blinding factors. We need for 
I created a temporary workaround in #14, but unfortunately, that's not entirely enough. We also need tweaks for other functions like computing the last blinding factor. This currently has functions that assume `SecretKey` type blinding factors that are not compatible with explicit assets.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -37,6 +37,7 @@
 #![cfg_attr(all(not(test), not(feature = "std")), no_std)]
 #![cfg_attr(all(test, feature = "unstable"), feature(test))]
 
+#[macro_use]
 pub extern crate secp256k1_zkp_sys;
 pub use secp256k1_zkp_sys as ffi;
 
@@ -95,6 +96,10 @@ pub enum Error {
     InvalidRangeProof,
     /// Bad generator
     InvalidGenerator,
+    /// Tweak must of len 32
+    InvalidTweakLength,
+    /// Tweak must be less than secp curve order
+    TweakOutOfBounds,
     /// Given bytes don't represent a valid adaptor signature
     InvalidEcdsaAdaptorSignature,
     /// Failed to decrypt an adaptor signature because of an internal error within `libsecp256k1-zkp`
@@ -120,6 +125,8 @@ impl fmt::Display for Error {
             Error::CannotRecoverAdaptorSecret => "failed to recover adaptor secret",
             Error::CannotVerifyAdaptorSignature => "failed to verify adaptor signature",
             Error::Upstream(inner) => return write!(f, "{}", inner),
+            Error::InvalidTweakLength => "Tweak must of size 32",
+            Error::TweakOutOfBounds => "Tweak must be less than secp curve order",
         };
 
         f.write_str(str)
--- a/src/zkp/generator.rs
+++ b/src/zkp/generator.rs
@@ -1,7 +1,111 @@
 use core::{fmt, str};
-use ffi;
-use {constants, from_hex, Error, Secp256k1, SecretKey, Signing, Tag};
+use ffi::{self, CPtr};
+#[cfg(feature = "rand")]
+use rand::Rng;
+use {constants, from_hex, Error, Secp256k1, Signing, Tag};
 
+/// Represents a blinding factor/Tweak on secp256k1 curve
+///
+/// Contrary to a [`crate::SecretKey`], the value 0 is also a valid tweak.
+/// Values outside secp curve order are invalid tweaks.
+#[derive(Default, Hash)]
+pub struct Tweak([u8; constants::SECRET_KEY_SIZE]);
+impl_array_newtype!(Tweak, u8, constants::SECRET_KEY_SIZE);
+
+/// The zero Tweak
+pub const ZERO_TWEAK: Tweak = Tweak([
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+]);
+
+impl fmt::Debug for Tweak {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "Tweak(")?;
+        for i in self[..].iter().cloned() {
+            write!(f, "{:02x}", i)?;
+        }
+        write!(f, ")")
+    }
+}
+
+impl fmt::LowerHex for Tweak {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        for ch in &self.0[..] {
+            write!(f, "{:02x}", *ch)?;
+        }
+        Ok(())
+    }
+}
+
+impl fmt::Display for Tweak {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        fmt::LowerHex::fmt(self, f)
+    }
+}
+
+impl str::FromStr for Tweak {
+    type Err = Error;
+    fn from_str(s: &str) -> Result<Tweak, Error> {
+        let mut res = [0; constants::SECRET_KEY_SIZE];
+        match from_hex(s, &mut res) {
+            Ok(constants::SECRET_KEY_SIZE) => Tweak::from_inner(res),
+            _ => Err(Error::InvalidTweakLength),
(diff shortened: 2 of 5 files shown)


src/zkp/generator.rs
```diff
@@ -1,7 +1,79 @@
...
+use rand::Rng;
+use {constants, from_hex, Error, Secp256k1, Signing, Tag};
+/// Represents a blinding factor/Tweak on secp256k1 curve
+///
+/// This is same as [`crate::SecretKey`], but is allowed to take
+/// zero value and values outside of secp curve order
+pub struct Tweak([u8; constants::SECRET_KEY_SIZE]);
+impl_array_newtype!(Tweak, u8, constants::SECRET_KEY_SIZE);
+
+/// The number 1 encoded as a secret key
```
thomaseizinger, line 13: This doc-string is diverging from the implementation.

src/zkp/generator.rs
```diff
@@ -1,7 +1,79 @@
 use core::{fmt, str};
 use ffi;
-use {constants, from_hex, Error, Secp256k1, SecretKey, Signing, Tag};
+#[cfg(feature = "rand")]
+use rand::Rng;
+use {constants, from_hex, Error, Secp256k1, Signing, Tag};
+/// Represents a blinding factor/Tweak on secp256k1 curve
+///
+/// This is same as [`crate::SecretKey`], but is allowed to take
+/// zero value and values outside of secp curve order
```
thomaseizinger, line 9: Given that the definition of a `SecretKey` is that it cannot be zero and must be within the curve order, the only thing that this shares with a `SecretKey` is the length right?

What do you think about expressing this as:

```suggestion
/// Represents a blinding factor/Tweak on secp256k1 curve
///
/// Contrary to a [`crate::SecretKey`], any 32 bytes are a valid tweak. This includes `0` and values outside of the secp curve order.
```

src/zkp/generator.rs
```diff
@@ -1,7 +1,79 @@
 use core::{fmt, str};
 use ffi;
-use {constants, from_hex, Error, Secp256k1, SecretKey, Signing, Tag};
+#[cfg(feature = "rand")]
+use rand::Rng;
+use {constants, from_hex, Error, Secp256k1, Signing, Tag};
+/// Represents a blinding factor/Tweak on secp256k1 curve
```
thomaseizinger, line 6: Do you mind adding a newline in between the imports and the docs here?

src/zkp/generator.rs
```diff
@@ -1,7 +1,79 @@
...
+            _ => Err(Error::InvalidTweak),
+        }
+    }
+}
+
+impl Tweak {
+    /// Generate a new random Tweak
+    #[cfg(feature = "rand")]
+    pub fn new<R: Rng + ?Sized>(rng: &mut R) -> Tweak {
+        let mut ret = [0u8; 32];
```
thomaseizinger, line 58: Probably best to use the constant everywhere.

```suggestion
        let mut ret = [0u8; constants::SECRET_KEY_SIZE];
```

src/zkp/generator.rs
```diff
@@ -68,8 +136,7 @@ impl Generator {
     }
 
     /// Creates a new unblinded [`Generator`] by from [`Tag`]
-    /// The [Generator::new_blinded] cannot be used for zero blind because
-    /// zero value is not permitted for [`SecretKey`] type.
+    /// Same as using zero [`Tweak`] with [`Generator::new_blinded`]
```
thomaseizinger, line 139: Shall we implement this then by delegating to `new_blinded` with `ZERO_TWEAK`?

src/lib.rs
```diff
@@ -120,6 +122,7 @@ impl fmt::Display for Error {
             Error::CannotRecoverAdaptorSecret => "failed to recover adaptor secret",
             Error::CannotVerifyAdaptorSignature => "failed to verify adaptor signature",
             Error::Upstream(inner) => return write!(f, "{}", inner),
+            Error::InvalidTweak => "Tweak must of size 32",
```
thomaseizinger, line 125: I am a bit worried that being more specific in the `Display` impl compared to the variant name is going to cause headaches down the line where someone reuses `Error::InvalidTweak` without reading what it will print as.

Can we rename this variant to be `InvalidTweakLength`? I think that might also make handling this further down the line easier because that is often something that can be guaranteed and matching / unwrapping a specific error is easier compared to something as generic as `InvalidTweak`.

src/zkp/generator.rs
thomaseizinger: Shall we also add a `From<[u8; 32]>` implementation?

thomaseizinger (commented):
LGTM. A few minor suggestions :)

sanket1729:
@thomaseizinger, thanks for the prompt review. I agree with all of your nits, will address those shortly. 

I think I misunderstood the `blind32` parameter in libsecp256k1, it returns 0 even if the blind is out of range. So, this special case struct is only needed for `SecretKey` is 0. 

It is possible to go the `Option<SecretKey>` route for having `None` as the zero blinding key. But that would be difficult to use upstream and could bloat the code unnecessarily. I will update the current code to reject out-of-order curve points.

thomaseizinger:
> It is possible to go the `Option<SecretKey>` route for having `None` as the zero blinding key. But that would be difficult to use upstream and could bloat the code unnecessarily. I will update the current code to reject out-of-order curve points.

Agreed that this feels wrong. Better have a dedicated type that allows 0 as its value.

sanket1729 (commented):
Addressed the suggestions and added a commit to delegate the implementation of `Generator::new_unblinded` and `PedersenCommitment::new_unblinded`

src/zkp/generator.rs
```diff
@@ -1,7 +1,115 @@
...
+        match data.len() {
+            constants::SECRET_KEY_SIZE => {
+                let mut ret = [0; constants::SECRET_KEY_SIZE];
+                unsafe {
+                    if ffi::secp256k1_ec_seckey_verify(
+                        ffi::secp256k1_context_no_precomp,
+                        data.as_ref().as_c_ptr(),
+                    ) == 0
+                    {
+                        if data.iter().all(|x| *x == 0) {
```
sanket1729, line 77: I am not sure if this is a constant-time crypto concern. A special case of [0; 32].

sanket1729:
Musing whether I should rename `Tweak` to `Scalar`

thomaseizinger (approved):
> Musing whether I should rename `Tweak` to `Scalar`

I'll leave that up to you :)
It would be nice to eventually have better support for arbitrary scalar computations so we could use `rust-secp256k1-zkp` as a backend for @LLFourn's https://github.com/LLFourn/secp256kfun.

src/zkp/generator.rs
sanket1729: Here is my rationale for the above change. 
Assuming we care about constant time, we only care about constant time properties for valid keys. With the above code, we guarantee the constant time property for all valid tweaks but zero. 
Although this is not ideal, this suffices in elements use-case where zero is usually as explicit commitment. 

Checking this upfront might however leak information about valid Tweaks (based on the position of first 1). I can however move the code afterward with a short circuit return the valid non-zero tweak in the unsafe block. 

Or I may just be overthinking this :man_shrugging: .

thomaseizinger: Thanks for clarifying, this does make sense! Happy to leave the code where it is.

Do you mind adding this information to the commit message?

apoelstra: I think this is fine as-is.

Strictly speaking 0 is as good a tweak as any other value, and this is a violation of constant-timeness. Having said this, the only way to get 0 in practice is to do so deliberately -- "standard" use of CT means that these values will always be either uniformly random or forced by the balancing equation, and in both cases will be nonzero with overwhelming probability.

So if somebody puts a 0 in here, they must be using some weird algorithm, and use of this weird algorithm will be detectable by 3rd parties by directly observing the commitments (and grinding through possible values). So sidechannel leaks would be the least of their concerns.

Fixing this "properly" would require adding functionality upstream in libsecp-zkp or even libsecp, which does not seem worthwhile for this edge case.

apoelstra (approved):
ack 0fd14bc5f123c7d9ef4d9e957b9125c982f4db8b

PR #17, opened 2021-04-26
Add derives for surjectionproofs

sanket1729:


thomaseizinger:
That doesn't work on 1.29 unfortunately :(

We will need a fairly recent rustc for this because it requires min-const-generics AFAIK.

sanket1729:
Is there any strong reason to not do a simple manual implementation for these traits?

thomaseizinger:
> Is there any strong reason to not do a simple manual implementation for these traits?

As in, loop through the array and check each element?

For an 8000 element array, this might be expensive although I haven't conducted any experiments. Happy to merge a manual implementation if you need these traits!

sanket1729:
Moved to #16

PR #22, opened 2021-05-04
Update secp256k1-zkp sys crate

sanket1729:
Update to rev f3708a1ecb445b1b05a0f8fcd1da6a88f83d89c4 on master


apoelstra (approved):
utack, confirmed that vendoring was done correctly

PR #23, opened 2021-07-12
Implement AsRef for EcdsaAdaptorSignature

Tibo-lg:
While doing some serialization work I noticed that I had omitted to implement something to get an `EcdsaAdaptorSignature` as a slice. As implementing `AsRef` seems to be what is done in rust-secp256k1 I went with that but let me know if there is another preferred way.

### DIFF
--- a/src/zkp/ecdsa_adaptor.rs
+++ b/src/zkp/ecdsa_adaptor.rs
@@ -61,6 +61,12 @@ impl CPtr for EcdsaAdaptorSignature {
     }
 }
 
+impl AsRef<[u8]> for EcdsaAdaptorSignature {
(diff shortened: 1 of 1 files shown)


thomaseizinger (approved):
Works for me!

There is formatting to hex via `fmt::Display` in case your serialization needs are human-facing.

Tibo-lg:
> There is formatting to hex via fmt::Display in case your serialization needs are human-facing.

It's really to write to a buffer that is sent on wire but thanks :)

jonasnick (commented):
ACK 687bb98f3db50617f18fbc366a83504b65e983d1

PR #24, opened 2021-07-15
Build and test wasm on MacOS in CI

rishflab:
See https://github.com/rust-bitcoin/rust-secp256k1/pull/254 for more
details


.github/workflows/rust.yml
```diff
@@ -29,28 +29,35 @@ jobs:
...
+      - name: Set CC and AR vars for MacOS
+        if: contains(matrix.os, 'macos')
+        run: |
+          echo "CC=/usr/local/opt/llvm/bin/clang" >> $GITHUB_ENV
+          echo "AR=/usr/local/opt/llvm/bin/llvm-ar" >> $GITHUB_ENV
       - name: Building docs
         env:
           DO_DOCS: true
         run: ./contrib/test.sh
-      - name: Install wasm-pack
```
thomaseizinger, line 52: Why are we deleting this?

It doesn't actually run anything now: https://github.com/ElementsProject/rust-secp256k1-zkp/runs/3075531208#step:6:68

Also, the script seems to be flawed because it doesn't fail the build if it is not succeeding 😅

Looking at the script, I also saw that this is already setting the CC variable so we either want to remove that or what we do in the CI jobs is pointless.

thomaseizinger (changes requested):

.github/workflows/rust.yml
rishflab: I guess the original is broken as well https://github.com/rust-bitcoin/rust-secp256k1/pull/254. We also don't install llvm? did this every work in CI?

bonomat: You still need wasm-pack to run the wasm tests :)

rishflab: added on line 60 of 0b4c3dbb34b2ec8fd118aa2a2ea46d6034e26cee

thomaseizinger: > We also don't install llvm?

I think llvm just comes pre-installed as @bonomat discovered [here](https://github.com/comit-network/waves/pull/170).

Cargo.toml
```diff
@@ -45,3 +45,6 @@ bitcoin_hashes = "0.9"
 [target.wasm32-unknown-unknown.dev-dependencies]
 wasm-bindgen-test = "0.3"
 rand = { version = "0.6", features = ["wasm-bindgen"] }
+
+[lib]
+crate-type = ["cdylib", "rlib"]
```
thomaseizinger, line 50: I did some testing locally where I deleted this and it doesn't seem to have an impact?

I can build and run the tests on wasm without this being specified.

After doing some digging, I found [this](https://stackoverflow.com/questions/56227766/why-must-a-wasm-library-in-rust-set-the-crate-type-to-cdylib). I _think_ we only need to set this for the crate that actually uses `wasm_bindgen` to export functions from a WASM module.

Unless someone wants to directly use the functions provided in this library from JS (i.e. without creating a new crate and declaring a dependency on this one), setting this here is just noise and doesn't actually do anything useful I believe.

contrib/test.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/sh -ex
 
+set -e
```
thomaseizinger, line 3: Being confused about how this script ever worked, I did some digging and saw that this script is actually launched with `-e`. See the shebang line :)

This makes me wonder, this actually change something? Did you test whether or not this makes a difference?

.github/workflows/rust.yml
thomaseizinger: See the CI run in this repository: https://github.com/ElementsProject/rust-secp256k1-zkp/pull/24/checks?check_run_id=3082712519#step:6:4

thomaseizinger (changes requested):

.github/workflows/rust.yml
```diff
@@ -29,43 +29,46 @@ jobs:
...
       - name: Install wasm-pack
         run: curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh
+      - name: Install llvm
+        if: matrix.os == 'macos-latest'
+        run: brew install llvm
       - name: Running WASM build
         env:
           DO_WASM: true
+          WASM_CC: ${{ matrix.cc }}
+          WASM_AR: ${{ matrix.ar }}
```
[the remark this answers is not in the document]
rishflab: agreed, I was unsure whether you could set env variables like that because of how `DO_WASM` is handled

.github/workflows/rust.yml
rishflab: It seems to be specific to my machine (don't need set -e in the mac os CI):

https://github.com/rishflab/rust-secp256k1-zkp/pull/1/commits/e1e86ffeac0b0324175bfd421e10c4bf1f2d54d6
https://github.com/rishflab/rust-secp256k1-zkp/pull/1/checks?check_run_id=3099601565

https://github.com/rishflab/rust-secp256k1-zkp/pull/1/commits/298c2944d57a41dfeb24e132e88d08aca4f8a49e
https://github.com/rishflab/rust-secp256k1-zkp/pull/1/checks?check_run_id=3099702619

Cargo.toml
rishflab: Did not pass CI without specifying crate-type

contrib/test.sh
rishflab: Made no difference, wasm-pack errors are swallowed on Mac OS ci

thomaseizinger (approved):
Looks good from my side! Thanks for sticking with this @rishflab :)

@jonasnick / @sanket1729 Do you mind having a look at this?

jonasnick (commented):
I don't know anything about WASM, but this PR is fine from my POV since the other tests aren't affected. Why do we remove the `build_wasm` task?

rishflab:
> I don't know anything about WASM, but this PR is fine from my POV since the other tests aren't affected. Why do we remove the `build_wasm` task?

The `contrib/test.sh` script does the build anyway (through wasm-pack) so this step is kind of redundant?

jonasnick:
Sorry, missed your answer. Ok.
ACK 149d68657fa5135f0eacf10ff3d7e391a5410f2a

jonasnick:
Thank you for the PR.

PR #25, opened 2021-07-20
Errors are swallowed in CI

rishflab:
This branch is fails intentionally on the `wasm-pack build` step. The error is swallowed in CI and it shows green.

#26

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -29,13 +29,21 @@ jobs:
           DO_BENCH: true
         run: ./contrib/test.sh
 
-  wasm_pack:
+  wasm:
     name: Stable - Docs / WebAssembly Build
-    runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


jonasnick (commented):
This is normal bash behavior. You can test this with
```
#!/bin/sh  -e
false && true
echo "foo"
```

I think you can just remove `&&`.

rishflab:
> This is normal bash behavior. You can test this with
> 
> ```
> #!/bin/sh  -e
> false && true
> echo "foo"
> ```
> 
> I think you can just remove `&&`.

ahhh
thanks!

jonasnick:
Looks good to me. Needs rebase and undrafting.

jonasnick (approved):
ACK 9a31a51f07aac2198b02e2c4aed2fe1b9d1c8d7f

PR #27, opened 2021-08-13
secp256k1-zkp-sys: Fix typo for SHA256 tagged hash

GeneFerneau:
Correct function call for MuSig SHA256 tagged hash initialization
Corrected call from Schnorrsig SHA256 tagged hash

### DIFF
--- a/secp256k1-zkp-sys/depend/secp256k1/src/modules/musig/main_impl.h
+++ b/secp256k1-zkp-sys/depend/secp256k1/src/modules/musig/main_impl.h
@@ -457,7 +457,7 @@ static void rustsecp256k1zkp_v0_4_0_musig_compute_messagehash(const rustsecp256k
 
     VERIFY_CHECK(session->round >= 2);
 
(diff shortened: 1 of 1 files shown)


secp256k1-zkp-sys/depend/secp256k1/src/modules/musig/main_impl.h
```diff
@@ -457,7 +457,7 @@ static void rustsecp256k1zkp_v0_4_0_musig_compute_messagehash(const rustsecp256k
 
     VERIFY_CHECK(session->round >= 2);
 
-    rustsecp256k1zkp_v0_4_0_schnorrsig_sha256_tagged(&sha);
+    rustsecp256k1zkp_v0_4_0_musig_sha256_init_tagged(&sha);
```
GeneFerneau, line 460: Not sure if the original was intended. If it was, I can expose `rustsecp256k1zkp_v0_4_0_schnorrsig_sha256_tagged` in `secp256k1_schnorrsig.h`, and include the header.

PR #30, opened 2021-09-08
Remove explicit bitcoin_hashes dependency

thomaseizinger:
Instead of depending in bitcoin_hashes explicitly, we can just re-export the module from `secp256k1`. This should simplify future version upgrades.


jonasnick (commented):
LGTM, but we need to replace the `hashes` feature in `test.sh` with `bitcoin_hashes`.

jonasnick (approved):
ACK e5d82e728a0bcc1d81b0824e87e70efcf309b9b0

PR #34, opened 2021-09-14
Implement serde for EcdsaAdaptorSignature

luckysori:


src/zkp/rangeproof.rs
```diff
@@ -207,6 +208,17 @@ impl ::core::fmt::Display for RangeProof {
     }
 }
 
+impl std::str::FromStr for RangeProof {
+    type Err = Error;
+    fn from_str(s: &str) -> Result<RangeProof, Error> {
+        let mut res = Vec::new();
```
thomaseizinger, line 214: Could initialize this with `Vec::with_capacity(s.len() / 2)` to avoid repeated allocations as the string is decoded.

src/zkp/surjection_proof.rs
```diff
@@ -187,6 +188,17 @@ impl ::core::fmt::Display for SurjectionProof {
...
+    fn from_str(s: &str) -> Result<SurjectionProof, Error> {
+        let mut res = Vec::new();
+        match from_hex(s, &mut res) {
+            Ok(_) => SurjectionProof::from_slice(&res),
+            _ => Err(Error::InvalidSurjectionProof),
+        }
+    }
+}
+
 // TODO: Macrofy (de)serialization
```
thomaseizinger, line 202: I think this can be deleted now, I would consider the impl to be concise enough.

thomaseizinger (approved):
Neat, thank you!

CI is not passing but that should be easy to fix.

thomaseizinger (approved):
ACK 6ec06011f8cac3d7d6f83b1067648be2c4404c2e

jonasnick (commented):
Thank you. Looks good to me. Could you add a note to the CHANGELOG.md about implementing serde for ECDSA adaptor sigs?

thomaseizinger (approved):
ACK f089aea8a5f3004938782b633f0ad4b12ca186db

jonasnick (commented):
ACK f089aea8a5f3004938782b633f0ad4b12ca186db

PR #36, opened 2021-10-04
Add whitelist support

stevenroose:
Open for review, working on unit tests.

### DIFF
--- a/secp256k1-zkp-sys/src/zkp.rs
+++ b/secp256k1-zkp-sys/src/zkp.rs
@@ -1,9 +1,12 @@
 use core::{fmt, hash};
-use {types::*, Context, PublicKey, Signature};
+use {types::*, Context, NonceFn, PublicKey, Signature};
 
 /// Rangeproof maximum length
 pub const RANGEPROOF_MAX_LENGTH: size_t = 5134;
 pub const ECDSA_ADAPTOR_SIGNATURE_LENGTH: size_t = 162;
+
+/// The maximum number of whitelist keys.
+pub const WHITELIST_MAX_N_KEYS: size_t = 255;
 
 extern "C" {
     #[cfg_attr(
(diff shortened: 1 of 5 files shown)


src/zkp/whitelist.rs
```diff
@@ -0,0 +1,470 @@
...
+        let keys_online = (0..n_keys)
+            .map(|_| SECP256K1.generate_keypair(&mut rng))
+            .collect::<Vec<_>>();
+        let keys_offline = (0..n_keys)
+            .map(|_| SECP256K1.generate_keypair(&mut rng))
+            .collect::<Vec<_>>();
+
+        // extract the pubkeys to serve as PAK list
+        let pak_online = keys_online.iter().map(|p| p.1).collect::<Vec<_>>();
+        let pak_offline = keys_offline.iter().map(|p| p.1).collect::<Vec<_>>();
```
thomaseizinger, line 241: [`unzip`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.unzip) might be useful here.

src/zkp/whitelist.rs
```diff
@@ -0,0 +1,470 @@
...
+        };
+        if ret != 1 {
+            return Err(Error::InvalidWhitelistSignature);
+        }
+
+        Ok(WhitelistSignature(sig))
+    }
+
+    /// Create a new whitelist ring signature for the given PAK list and whitelist key.
+    pub fn sign<C: Signing>(
```
thomaseizinger, line 70: Shall we name this `new` instead?

`WhitelistSignature::sign` repeats the `sign` part.

secp256k1-zkp-sys/src/zkp.rs
```diff
@@ -476,6 +532,39 @@ impl hash::Hash for PedersenCommitment {
...
+impl Eq for WhitelistSignature {}
+
+impl Default for WhitelistSignature {
+    fn default() -> WhitelistSignature {
+        WhitelistSignature {
+            n_keys: 0,
+            data: [0; 32 * (1 + WHITELIST_MAX_N_KEYS)],
+        }
+    }
+}
```
thomaseizinger, line 566: I remember that there was a discussion in the past that these initializers might better be marked `unsafe`. Not sure if that applies here as well.

src/zkp/whitelist.rs
```diff
@@ -0,0 +1,470 @@
...
+    }
+}
+
+#[cfg(all(test, feature = "global-context"))]
+mod tests {
+    use super::*;
+    use rand::thread_rng;
+    use SECP256K1;
+
+    fn test_whitelist_proof_roundtrip(n_keys: usize) {
```
thomaseizinger, line 230: Would prefer to have helpers defined after their usage.

src/zkp/whitelist.rs
```diff
@@ -0,0 +1,470 @@
...
+use core::ptr;
+
+use ffi::CPtr;
+#[cfg(feature = "std")]
+use from_hex;
+use {ffi, secp256k1, Error, PublicKey, Secp256k1, SecretKey, Signing, Verification};
+
+/// A whitelist ring signature.
+#[derive(Clone, PartialEq, Eq, Hash)]
+#[repr(transparent)]
```
thomaseizinger, line 17: Does this `repr` allow a cast via `as` from `ffi::WhitelistSignature`? In that case, I think we should not have `Default` on `ffi::WhitelistSignature` but instead have an unsafe constructor.

Alternatively, we can remove `repr` here. What do we need it for?

src/zkp/whitelist.rs
thomaseizinger: For reference, `SurjectionProof` for example uses `new`: https://github.com/ElementsProject/rust-secp256k1-zkp/blob/0dba4c033a929164dd21c488d2a5a9790919f65c/src/zkp/surjection_proof.rs#L21-L31

src/zkp/whitelist.rs
stevenroose: Fancy, updated.

src/zkp/whitelist.rs
thomaseizinger: Well, it is a bit of a C89-ism if you ask me :)

The way I think about it usually is: When I look at an entire file / module, I want to see the most important things first (usually public API) or in the case of the `tests` module the tests.

While I am reading through the tests / top-level functions, I may come across a function I've not seen before. Assuming a reasonably descriptive name, I can continue reading (and understanding) the original function and thereby follow along the steps of what it does better without having to context-switch to the details of how that functions achieves its job.

If all of the helper functions are defined before they are used, reading the file top to bottom leaves me with a lot of code that is initially unrelated to each other and only once I get to the piece where it is used, it finally comes together.

It is a personal preference but I find the former to be easier :)

In this case however, it is debatable as literally the entire test is a delegation to another function so it might as well be defined at the top :D

src/zkp/whitelist.rs
thomaseizinger: I would prefer `sign` if we are calling it on whatever we are signing. Like `message.sign(key)`. In this case, that is not really possible from what I can see on the parameter list.

I am just co-maintainer here so happy to go with whatever as long as we are consistent :)

stevenroose: Changed.

PR #38, opened 2021-10-20
whitelist: Fix occasional overflow in unit test

stevenroose:
We bumped into this in master and will happen once every 255 runs on average: https://github.com/ElementsProject/rust-secp256k1-zkp/runs/3951901266?check_suite_focus=true

Just me forgetting that Rust is typed :|


jonasnick (approved):
utACK 02f16667fb6dafcba5b96b6dd2e6723e9ff80f2c

PR #41, opened 2021-10-22
Prepare for release 0.5.0

### LINKED ISSUE #40: New release
I would really love to have a new release that includes #39 so that I could have my project completely free of github dependencies and make a proper release of it. I also depend on the added serde serialization for `EcdsaAdaptorSignature`. I think that those two changes are non breaking and so I'm happy to prepare  a branch with only those two for a 0.4.1 release if people think it's a good way to go. I'm also happy with a 0.5.0 release based on master if people prefer that (as I believe the two renames are breaking).

Let me know if there is anything I could help with to make that happen.

jonasnick:
Fixes #40

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Unreleased
 
+# 0.5.0 - 2021-10-22
+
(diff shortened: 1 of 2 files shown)


thomaseizinger (approved):
ACK 582a5a0408e775fc38d5e7475d70c693750018ac

PR #42, opened 2021-11-04
Prepare for -sys release 0.5.0

jonasnick:
I messed up #41 by forgetting to also bump -sys.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,7 @@ use-rand = ["rand", "secp256k1/rand"]
 
 [dependencies]
 secp256k1 = "0.20.0"
-secp256k1-zkp-sys = { version = "0.4.0", default-features = false, path = "./secp256k1-zkp-sys" }
+secp256k1-zkp-sys = { version = "0.5.0", default-features = false, path = "./secp256k1-zkp-sys" }
 rand = { version = "0.6", default-features = false, optional = true }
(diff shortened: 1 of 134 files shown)


jonasnick:
Fwiw I also tried to update the vendored secp-zkp library, but that results in test failure which under valgrind shows that something is going very wrong memory wise and valgrind is pointing to something being wrong with the context. My hypothesis is that the preallocated context size changed in secp-zkp, but since we're reusing the `Context` from rust-secp, which is behind in C-libsecp the `Context` doesn't have the updated preallocated context size.

Tibo-lg:
Just trying to be useful, but I can confirm that tests fail with latest secp-zkp. And I get the exact same result using the previously used commit (f3708a1ecb445b1b05a0f8fcd1da6a88f83d89c4).

apoelstra:
We can update the library in rust-secp to match rust-secp-zkp since we are about to do a major release anyway.

@jonasnick suggested to me in another channel that he was working on this, so I'll leave that to him.

thomaseizinger:
@jonasnick Any chance we can get a release out without the changes to the vendored C library? I would like to get rid of Git dependencies :)

jonasnick (commented):
In order to get a release out we need to merge this PR, but no one approved it so far.

Tibo-lg (approved):
ack cdba7dc7dd83200f44fc301f3de86c9fe9a6b4e9

PR #43, opened 2022-01-07
Upgrade to secp256k1 0.21

### LINKED ISSUE #32: Upgrade at least to `rust-secp256k1` 0.21
The 0.21 release will include https://github.com/rust-bitcoin/rust-secp256k1/pull/326 which is required for interoperability with the `ThirtyTwoByteHash` trait. In order to use `rust-secp256k1-zkp` in a project together with `rust-secp256k1` 0.21, we will need to bump the version we are depending on and release a new version as well.

thomaseizinger:
Resolves #32.


thomaseizinger:
@apoelstra Any idea why the tests are failing? I've read something about an endianness issue related to the latest `secp256k1` release but can't find the comment now. Perhaps this is also an instance of that.

apoelstra:
@thomaseizinger I don't think it's about endianness, this looks like a memory issue, and I think it's because upstream's context object is now much smaller than the secp-zkp one, and this library treats them as being the same thing.

I think updating the embedded version of libsecp-zkp to one after https://github.com/ElementsProject/secp256k1-zkp/pull/159 should make things work, but I'm not sure, I'm not totally on top of the relative state of all these libraries.

thomaseizinger:
> @thomaseizinger I don't think it's about endianness, this looks like a memory issue, and I think it's because upstream's context object is now much smaller than the secp-zkp one, and this library treats them as being the same thing.
> 
> I think updating the embedded version of libsecp-zkp to one after [ElementsProject/secp256k1-zkp#159](https://github.com/ElementsProject/secp256k1-zkp/pull/159) should make things work, but I'm not sure, I'm not totally on top of the relative state of all these libraries.

Thanks for the pointer, I will try update the embedded lib. Totally forgot that we treating those contexts the same :)

apoelstra:
Yeah, we are really playing with fire there.

Maybe now that we have fully static contexts, we should stop using the upstream contents entirely and just create dummy contexts inside this lib's functions.

apoelstra:
Sorry, for close, trackpad user failure.

thomaseizinger:
> Yeah, we are really playing with fire there.
> 
> Maybe now that we have fully static contexts, we should stop using the upstream contents entirely and just create dummy contexts inside this lib's functions.

Upstream being `rust-secp256k1` here? Is the plan for `rust-secp` to drop the context arg entirely?

apoelstra:
@thomaseizinger see discussion in https://github.com/rust-bitcoin/rust-secp256k1/issues/343

The plan isn't totally clear yet.

thomaseizinger:
Closing in favor of https://github.com/ElementsProject/rust-secp256k1-zkp/pull/46.

PR #44, opened 2022-01-26
Fix build errors

sanket1729:
The upstream impl_array_newtype now implements a `core::hash::Hash`. This
caused a breaking changed here :( despite it being a minor release.

PR #45, opened 2022-02-07
Remove `Hash` implementations for `zkp` types

luckysori:
A breaking change was inadvertently introduced with the release of `rust-secp256k1-sys:0.4.2`. 

Anyone who had used `impl_array_newtype` for type `T` and had also implemented `hash::Hash` on `T` before version `0.4.2` would now come across a "conflicting implementations error" on version `0.4.2`.


luckysori:
Oops, didn't see #44. Nothing to see here.

PR #48, opened 2022-03-21
Musig2 support

sanket1729:
This was taken by #29 , but significantly changed. In particular, I was super strict about adding `unreachable!` to make the APIs not return a result when possible. Each API has a doc test with guidelines of using it. 

Things to pay special attention to:
- Cross-check the use of `unreachable!` in the FFI calls to make sure they cannot panic for any other reasons.
- Review the documentation about the usage of `session_id` and nonce re-use.

Edit: I wanted to add a extra-rand with time of day example, but doing so in the current MSRV is non-trivial. Will add an example after MSRV update

### DIFF
--- a/secp256k1-zkp-sys/src/zkp.rs
+++ b/secp256k1-zkp-sys/src/zkp.rs
@@ -1,8 +1,6 @@
-use core::{
-    fmt,
-    hash::{self, Hash},
-};
-use {types::*, Context, PublicKey, Signature};
+use core::{fmt, hash};
+use zkp::hash::Hash;
+use {types::*, Context, Keypair, PublicKey, Signature, XOnlyPublicKey};
 
 /// Rangeproof maximum length
 pub const RANGEPROOF_MAX_LENGTH: size_t = 5134;
@@ -391,6 +389,223 @@ extern "C" {
         n_keys: size_t,
         sub_pubkey: *const PublicKey,
     ) -> c_int;
+
+    #[cfg_attr(
+        not(feature = "external-symbols"),
+        link_name = "rustsecp256k1zkp_v0_8_0_musig_pubnonce_parse"
+    )]
+    pub fn secp256k1_musig_pubnonce_parse(
+        cx: *const Context,
+        nonce: *mut MusigPubNonce,
+        in66: *const c_uchar,
+    ) -> c_int;
+
+    #[cfg_attr(
+        not(feature = "external-symbols"),
+        link_name = "rustsecp256k1zkp_v0_8_0_musig_pubnonce_serialize"
+    )]
+    pub fn secp256k1_musig_pubnonce_serialize(
+        cx: *const Context,
+        out66: *mut c_uchar,
+        nonce: *const MusigPubNonce,
+    ) -> c_int;
+
+    #[cfg_attr(
+        not(feature = "external-symbols"),
+        link_name = "rustsecp256k1zkp_v0_8_0_musig_aggnonce_parse"
+    )]
+    pub fn secp256k1_musig_aggnonce_parse(
+        cx: *const Context,
+        nonce: *mut MusigAggNonce,
+        in66: *const c_uchar,
+    ) -> c_int;
+
+    #[cfg_attr(
+        not(feature = "external-symbols"),
+        link_name = "rustsecp256k1zkp_v0_8_0_musig_aggnonce_serialize"
+    )]
+    pub fn secp256k1_musig_aggnonce_serialize(
+        cx: *const Context,
+        out66: *mut c_uchar,
+        nonce: *const MusigAggNonce,
+    ) -> c_int;
+
+    #[cfg_attr(
+        not(feature = "external-symbols"),
+        link_name = "rustsecp256k1zkp_v0_8_0_musig_partial_sig_parse"
+    )]
+    pub fn secp256k1_musig_partial_sig_parse(
+        cx: *const Context,
+        sig: *mut MusigPartialSignature,
+        in32: *const c_uchar,
+    ) -> c_int;
+
+    #[cfg_attr(
+        not(feature = "external-symbols"),
+        link_name = "rustsecp256k1zkp_v0_8_0_musig_partial_sig_serialize"
+    )]
+    pub fn secp256k1_musig_partial_sig_serialize(
+        cx: *const Context,
+        out32: *mut c_uchar,
+        sig: *const MusigPartialSignature,
+    ) -> c_int;
+
+    #[cfg_attr(
+        not(feature = "external-symbols"),
+        link_name = "rustsecp256k1zkp_v0_8_0_musig_pubkey_agg"
+    )]
+    pub fn secp256k1_musig_pubkey_agg(
+        cx: *const Context,
+        scratch: *mut ScratchSpace,
+        combined_pk: *mut XOnlyPublicKey,
+        keyagg_cache: *mut MusigKeyAggCache,
+        pubkeys: *const *const PublicKey,
+        n_pubkeys: size_t,
+    ) -> c_int;
+
+    #[cfg_attr(
+        not(feature = "external-symbols"),
+        link_name = "rustsecp256k1zkp_v0_8_0_musig_pubkey_get"
+    )]
+    pub fn secp256k1_musig_pubkey_get(
+        cx: *const Context,
+        agg_pk: *mut PublicKey,
+        keyagg_cache: *const MusigKeyAggCache,
+    ) -> c_int;
+
+    #[cfg_attr(
+        not(feature = "external-symbols"),
+        link_name = "rustsecp256k1zkp_v0_8_0_musig_pubkey_ec_tweak_add"
+    )]
+    pub fn secp256k1_musig_pubkey_ec_tweak_add(
+        cx: *const Context,
+        output_pubkey: *mut PublicKey,
+        keyagg_cache: *mut MusigKeyAggCache,
+        tweak32: *const c_uchar,
+    ) -> c_int;
+
+    #[cfg_attr(
+        not(feature = "external-symbols"),
+        link_name = "rustsecp256k1zkp_v0_8_0_musig_pubkey_xonly_tweak_add"
+    )]
+    pub fn secp256k1_musig_pubkey_xonly_tweak_add(
+        cx: *const Context,
+        output_pubkey: *mut PublicKey,
+        keyagg_cache: *mut MusigKeyAggCache,
+        tweak32: *const c_uchar,
+    ) -> c_int;
+
+    #[cfg_attr(
+        not(feature = "external-symbols"),
+        link_name = "rustsecp256k1zkp_v0_8_0_musig_nonce_gen"
+    )]
+    pub fn secp256k1_musig_nonce_gen(
+        cx: *const Context,
+        secnonce: *mut MusigSecNonce,
+        pubnonce: *mut MusigPubNonce,
+        session_id32: *const c_uchar,
+        seckey: *const c_uchar,
+        pubkey: *const PublicKey,
+        msg32: *const c_uchar,
+        keyagg_cache: *const MusigKeyAggCache,
+        extra_input32: *const c_uchar,
+    ) -> c_int;
+
+    #[cfg_attr(
+        not(feature = "external-symbols"),
+        link_name = "rustsecp256k1zkp_v0_8_0_musig_nonce_agg"
+    )]
+    pub fn secp256k1_musig_nonce_agg(
+        cx: *const Context,
+        aggnonce: *mut MusigAggNonce,
+        pubnonces: *const *const MusigPubNonce,
+        n_pubnonces: size_t,
+    ) -> c_int;
+
+    #[cfg_attr(
+        not(feature = "external-symbols"),
+        link_name = "rustsecp256k1zkp_v0_8_0_musig_nonce_process"
+    )]
+    pub fn secp256k1_musig_nonce_process(
+        cx: *const Context,
+        session: *mut MusigSession,
+        aggnonce: *const MusigAggNonce,
+        msg32: *const c_uchar,
+        keyagg_cache: *const MusigKeyAggCache,
+        adaptor: *const PublicKey,
+    ) -> c_int;
+
+    #[cfg_attr(
+        not(feature = "external-symbols"),
+        link_name = "rustsecp256k1zkp_v0_8_0_musig_partial_sign"
+    )]
+    pub fn secp256k1_musig_partial_sign(
+        cx: *const Context,
+        partial_sig: *mut MusigPartialSignature,
+        secnonce: *mut MusigSecNonce,
+        keypair: *const Keypair,
+        keyagg_cache: *const MusigKeyAggCache,
+        session: *const MusigSession,
+    ) -> c_int;
+
+    #[cfg_attr(
+        not(feature = "external-symbols"),
+        link_name = "rustsecp256k1zkp_v0_8_0_musig_partial_sig_verify"
+    )]
+    pub fn secp256k1_musig_partial_sig_verify(
+        cx: *const Context,
+        partial_sig: *const MusigPartialSignature,
+        pubnonce: *const MusigPubNonce,
+        pubkey: *const PublicKey,
+        keyagg_cache: *const MusigKeyAggCache,
+        session: *const MusigSession,
+    ) -> c_int;
+
+    #[cfg_attr(
+        not(feature = "external-symbols"),
+        link_name = "rustsecp256k1zkp_v0_8_0_musig_partial_sig_agg"
+    )]
+    pub fn secp256k1_musig_partial_sig_agg(
+        cx: *const Context,
+        sig64: *mut c_uchar,
+        session: *const MusigSession,
+        partial_sigs: *const *const MusigPartialSignature,
+        n_sigs: size_t,
+    ) -> c_int;
+
+    #[cfg_attr(
+        not(feature = "external-symbols"),
+        link_name = "rustsecp256k1zkp_v0_8_0_musig_nonce_parity"
+    )]
+    pub fn secp256k1_musig_nonce_parity(
+        cx: *const Context,
+        nonce_parity: *mut c_int,
+        session: *const MusigSession,
+    ) -> c_int;
+
+    #[cfg_attr(
+        not(feature = "external-symbols"),
+        link_name = "rustsecp256k1zkp_v0_8_0_musig_adapt"
+    )]
+    pub fn secp256k1_musig_adapt(
+        cx: *const Context,
+        sig64: *mut c_uchar,
+        pre_sig64: *const c_uchar,
+        sec_adaptor32: *const c_uchar,
+        nonce_parity: c_int,
+    ) -> c_int;
+
+    #[cfg_attr(
+        not(feature = "external-symbols"),
+        link_name = "rustsecp256k1zkp_v0_8_0_musig_extract_adaptor"
+    )]
+    pub fn secp256k1_musig_extract_adaptor(
+        cx: *const Context,
+        sec_adaptor32: *mut c_uchar,
+        sig64: *const c_uchar,
+        pre_sig64: *const c_uchar,
+        nonce_parity: c_int,
+    ) -> c_int;
 }
 
 #[repr(C)]
@@ -622,6 +837,14 @@ impl Default for EcdsaAdaptorSignature {
     }
 }
 
+impl PartialEq for EcdsaAdaptorSignature {
+    fn eq(&self, other: &Self) -> bool {
+        &self.0[..] == &other.0[..]
+    }
+}
+
+impl Eq for EcdsaAdaptorSignature {}
+
 impl EcdsaAdaptorSignature {
     /// Create a new (zeroed) ecdsa adaptor signature usable for the FFI interface
     pub fn new() -> Self {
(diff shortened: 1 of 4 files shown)


sanket1729:
Marked as ready for review.

sanket1729:
I am not an ASAN expert, but I am not sure why adding this test causes a stack use after a return error. I inspected the code when it clearly does not. 

```rust
    #[test]
    fn test_nonce_parsing() {
        let secp = Secp256k1::new();
        let sec_bytes = [1; 32];
        let sec_key = SecretKey::from_slice(&sec_bytes).unwrap();
        let keypair = KeyPair::from_secret_key(&secp, sec_key);
        let pub_key = XOnlyPublicKey::from_keypair(&keypair);

        let (key_agg_cache, _agg_pk) = MusigKeyAggCache::new(&secp, &[pub_key, pub_key]);
        let msg = Message::from_slice(&[3; 32]).unwrap();
        let session_id = [2; 32];
        let sec_key = SecretKey::from_slice(&[4; 32]).unwrap();
        let (_secnonce, pubnonce) = new_musig_nonce_pair(
            &secp,
            session_id,
            Some(&key_agg_cache),
            None,
            Some(msg),
            None,
        );
        let pubnonce_ser = pubnonce.serialize();
        let parsed_pubnonce = MusigPubNonce::from_slice(&pubnonce_ser).unwrap();

        assert_eq!(parsed_pubnonce, pubnonce);
    }
```

jonasnick (commented):
I have only read through half the code yet but looks nice so far.

jonasnick (commented):
First review pass done. I wonder if it is worth having so much redundancy in the doc examples versus just having a single example that shows an entire signing session.

src/zkp/musig.rs
```diff
@@ -0,0 +1,1733 @@
...
+    /// let (key_agg_cache, _agg_pk) = MusigKeyAggCache::new(&secp, &[pub_key1, pub_key2]);
+    /// // The session id must be sampled at random. Read documentation for more details.
+    /// let mut session_id = [0; 32];
+    /// thread_rng().fill_bytes(&mut session_id);
+    ///
+    /// // Generate the nonce for party with `keypair1`.
+    /// let sec_key = SecretKey::from_keypair(&keypair1);
+    /// let msg = Message::from_slice(&[3; 32]).unwrap();
+    /// let mut extra_rand = [0u8; 32];
+    /// thread_rng().fill_bytes(&mut extra_rand);
```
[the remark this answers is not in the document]
sanket1729: Will add an example here that uses time

src/zkp/musig.rs
```diff
@@ -0,0 +1,1733 @@
...
+            cache_ptr,
+            extra_ptr,
+        ) == 0
+        {
+            // Rust type system guarantees that
+            // - input secret key is valid
+            // - msg is 32 bytes
+            // - Key agg cache is valid
+            // - extra input is 32 bytes
+            unreachable!("All input arguments are valid well and well-typed")
```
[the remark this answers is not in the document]
sanket1729: yep, will add a dedicated error type.

src/zkp/musig.rs
```diff
@@ -0,0 +1,1733 @@
...
+        } else {
+            (sec_nonce, pub_nonce)
+        }
+    }
+}
+
+/// Opaque data structure that holds a partial MuSig signature.
+///
+/// Guaranteed to be 32 bytes in size. Serialized and parsed with
```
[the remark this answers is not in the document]
sanket1729: Will remove these size statements as they are not necessary.

src/zkp/musig.rs
```diff
@@ -0,0 +1,1733 @@
...
+    }
+}
+
+impl MusigPubNonce {
+    /// Serialize a MusigPubNonce
+    ///
+    /// # Errors:
+    ///
+    /// - InvalidMusigPartSigLen: If the signature is not 32 bytes
+    /// - InvalidMusigPartSig: If the signature is 32 bytes, but out of curve order
```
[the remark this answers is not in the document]
sanket1729: Good catch. This was meant for the below function. And is the result of erroneous copy paste.

src/zkp/musig.rs
```diff
@@ -0,0 +1,1733 @@
...
+            if ffi::secp256k1_musig_nonce_agg(
+                *secp.ctx(),
+                aggnonce.as_mut_ptr(),
+                nonce_ptrs.as_ptr(),
+                nonce_ptrs.len(),
+            ) == 0
+            {
+                // This can only crash if the individual nonces are invalid which is not possible is rust.
+                // Note that even if aggregate nonce is point at infinity, the musig spec sets it as `G`
+                unreachable!("Public key nonces are well-formed and valid in rust typesystem")
```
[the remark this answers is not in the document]
sanket1729: Good catch. I think it is okay to panic in this case.

src/zkp/musig.rs
```diff
@@ -0,0 +1,1733 @@
...
+            }
+        }
+    }
+
+    /// Deserialize a MusigAggNonce from byte slice
+    ///
+    /// # Errors:
+    ///
+    /// - InvalidMusigPartSigLen: If the slice is not 66 bytes
+    /// - InvalidMusigPartSig: If the byte slice is 66 bytes, but the [`MusigAggNonce`] is invalid
```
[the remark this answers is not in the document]
sanket1729: Sorry for the sloppy work here. Lots of copy-pasting

sanket1729 (commented):
Sorry for the delay @jonasnick. Addressed the reviews.

src/zkp/musig.rs
```diff
@@ -0,0 +1,1733 @@
...
+    /// # # [cfg(any(test, feature = "rand-std"))] {
+    /// # use secp256k1_zkp::rand::{thread_rng, RngCore};
+    /// # use secp256k1_zkp::{MusigKeyAggCache, Secp256k1, SecretKey, KeyPair, XOnlyPublicKey};
+    /// let secp = Secp256k1::new();
+    /// let keypair1 = KeyPair::new(&secp, &mut thread_rng());
+    /// let pub_key1 = XOnlyPublicKey::from_keypair(&keypair1);
+    /// let keypair2 = KeyPair::new(&secp, &mut thread_rng());
+    /// let pub_key2 = XOnlyPublicKey::from_keypair(&keypair2);
+    ///
+    /// let (_key_agg_cache, _agg_pk) = MusigKeyAggCache::new(&secp, &[pub_key1, pub_key2]);
```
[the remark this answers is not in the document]
jonasnick: I don't understand my suggestion anymore. I think I was mainly complaining about having to call something ominously named `MusigKeyAggCache::new` if you just want an aggregate key. I don't see how my suggestion improves the situation.

Tibo-lg:
Thanks for providing this! I played a bit with it and I'm curious about the reason for having separate `Error` enums. It feels a bit impractical to me as for example if I want to easily convert them to my library error type I need to implement `From` many times. Or maybe there is a specific reason behind this choice?

I also spotted quite some typos in the docs but not sure if it's the right time to be nitpicking (if it is let me know I can point them out).

Otherwise it's really neat :)!

sanket1729:
> Or maybe there is a specific reason behind this choice?

@Tibo-lg Indeed. IMO, the return type of the function should only return the corresponding error variants that are expected from the function. If we collect all of them into a single enum, it is would impossible for the caller to know which variants are reachable for the corresponding function. This is handy for callers that want to act on an error instead of propagating forward.

```rust
match key_agg_cache.pubkey_ec_tweak_add(&secp, tweak) {
	Ok(_) => {},
	Err(MusigTweakError::InvalidTweak) => panic!("not possible in bip32 tweaking"),
}
```
If there is a single variant with everything, the caller has no choice but to propagate it forward. Unfortunately, there is a tradeoff that the implementors that want to forward the error instead of dealing with it would have more code to deal with. But overall, I think the tradeoff is clear as it clearly expresses the conditions the function errors with and allows sanely dealing with errors. 

> I also spotted quite some typos in the docs but not sure if it's the right time to be nitpicking (if it is let me know I can point them out).

Corrections welcome.

Tibo-lg (commented):
Just some nits on the comments. Regarding the error approach, I don't have a strong opinion but I just feel if that's the direction the library is taking it would probably be nice to refactor the other errors to follow the same approach at some point.

> If there is a single variant with everything, the caller has no choice but to propagate it forward.

Maybe I didn't fully understand what you meant, but can't you just `if let` or match on the variant you're interested in acting upon? Something like:

```rust
let res = key_agg_cache.pubkey_ec_tweak_add(&secp, tweak);
if let Err(Error::MusigTweakError(MusigTweakError::InvalidTweak)) = res {
    panic!("not possible in bip32 tweaking");
}
res
```

jonasnick:
The MuSig implementation in libsecp256k1-zkp has significantly changed while it was updated to the BIP version 1. Therefore, we should update the libsecp256k1-zkp version we use here.

src/zkp/musig.rs
```diff
@@ -0,0 +1,1762 @@
...
+    }
+
+    /// Starts a signing session by generating a nonce
+    ///
+    /// This function outputs a secret nonce that will be required for signing and a
+    /// corresponding public nonce that is intended to be sent to other signers.
+    ///
+    /// MuSig differs from regular Schnorr signing in that implementers _must_ take
+    /// special care to not reuse a nonce. If you cannot provide a `sec_key`, `session_id`
+    /// UNIFORMLY RANDOM AND KEPT SECRET (even from other signers).
```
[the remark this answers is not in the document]
Ademan: I noticed this as well, the next line "Refer to..." should be the end of this sentence. An older version was that way https://github.com/BlockstreamResearch/rust-secp256k1-zkp/pull/48#discussion_r856537263 the fix is probably just to join them again.

src/zkp/musig.rs
```diff
@@ -0,0 +1,1588 @@
...
+    /// Refer to libsecp256k1-zkp documentation for additional considerations.
+    ///
+    /// Musig2 nonces can be precomputed without knowing the aggregate public key, the message to sign.
+    /// However, for maximal mis-use resistance, this API requires user to have already
+    /// have [`SecretKey`], [`Message`] and [`MusigKeyAggCache`]. See the `new_nonce_pair` method
```
Ademan, line 284: > have [`SecretKey`]

`nonce_gen()` appears to take a public key, not a secret key.

stevenroose:
> @[sanket1729](https://github.com/sanket1729) sanket1729 changed the title ~Comment out Wasm build on CI~ Musig2 support [Mar 21, 2022](https://github.com/BlockstreamResearch/rust-secp256k1-zkp/pull/48#event-6278612901)

Well that escalated quickly.. I'm super excited about this though, guys!

Ademan:
It would be nice to have serde serialization for pubnonces and partial signatures. (I may provide a patch for that)

ssantos21:
Exposing the inner bytes of `ffi::MusigSecNonce` and `ffi::MusigSession` would allow to serialize and store them.

`secp256k1-zkp-sys/src/zkp.rs`
```diff
 #[repr(C)]
 #[derive(Copy, Clone)]
-pub struct MusigSecNonce([c_uchar; MUSIG_SECNONCE_LEN]);
+pub struct MusigSecNonce(pub [c_uchar; MUSIG_SECNONCE_LEN]);
 impl_array_newtype!(MusigSecNonce, c_uchar, MUSIG_SECNONCE_LEN);
 impl_raw_debug!(MusigSecNonce);
 
@@ -1008,7 +1008,7 @@ impl MusigAggNonce {
 
 #[repr(C)]
 #[derive(Copy, Clone)]
-pub struct MusigSession([c_uchar; MUSIG_SESSION_LEN]);
+pub struct MusigSession(pub [c_uchar; MUSIG_SESSION_LEN]);
 impl_array_newtype!(MusigSession, c_uchar, MUSIG_SESSION_LEN);
 impl_raw_debug!(MusigSession);

```

`src/zkp/musig.rs`

```diff
 use crate::{Signing, Verification};
+use ffi::{MUSIG_SECNONCE_LEN, MUSIG_SESSION_LEN};
 use secp256k1::Parity;

impl MusigSecNonce {
     pub fn as_mut_ptr(&mut self) -> *mut ffi::MusigSecNonce {
         &mut self.0
     }
+
+    /// Function to return a copy of the internal array
+    pub fn serialize(&self) -> [u8; MUSIG_SECNONCE_LEN] {
+        let ffi::MusigSecNonce(array) = &self.0;
+        *array
+    }
+
+    /// Function to create a new MusigKeyAggCoef from a 32 byte array
+    pub fn from_slice(array: [u8; MUSIG_SECNONCE_LEN]) -> Self {
+        MusigSecNonce(ffi::MusigSecNonce(array))
+    }
 }

 /// A musig Singing session.
 #[derive(Debug, Clone, Copy, Eq, PartialEq)]
-pub struct MusigSession(ffi::MusigSession);
+pub struct MusigSession(pub ffi::MusigSession);
 
 impl CPtr for MusigSession {
     type Target = ffi::MusigSession;
@@ -1671,6 +1695,17 @@ impl MusigSession {
     pub fn as_mut_ptr(&mut self) -> *mut ffi::MusigSession {
         &mut self.0
     }
+
+    /// Function to return a copy of the internal array
+    pub fn serialize(&self) -> [u8; MUSIG_SESSION_LEN] {
+        let ffi::MusigSession(array) = &self.0;
+        *array
+    }
+
+    /// Function to create a new MusigKeyAggCoef from a 32 byte array
+    pub fn from_slice(array: [u8; MUSIG_SESSION_LEN]) -> Self {
+        MusigSession(ffi::MusigSession(array))
+    }
 }


```

secp256k1-zkp-sys/src/zkp.rs
```diff
@@ -363,6 +373,28 @@ extern "C" {
...
+        tweak32: *const c_uchar,
+    ) -> c_int;
+
+    #[cfg_attr(
+        not(feature = "external-symbols"),
+        link_name = "rustsecp256k1zkp_v0_8_0_musig_pubkey_xonly_tweak_add"
+    )]
+    pub fn secp256k1_musig_pubkey_xonly_tweak_add(
+        cx: *const Context,
+        output_pubkey: *mut XOnlyPublicKey,
```
ssantos21, line 393: `secp256k1_musig_pubkey_xonly_tweak_add` expects a `PublicKey`, not `XOnlyPublicKey`.
This is causing signature verification to fail.

```suggestion
        output_pubkey: *mut PublicKey,
```

src/zkp/musig.rs
```diff
@@ -0,0 +1,1475 @@
...
+    /// # }
+    /// ```
+    pub fn pubkey_xonly_tweak_add<C: Verification>(
+        &mut self,
+        secp: &Secp256k1<C>,
+        tweak: SecretKey,
+    ) -> Result<XOnlyPublicKey, MusigTweakErr> {
+        let cx = secp.ctx().as_ptr();
+        unsafe {
+            let mut out = XOnlyPublicKey::from(ffi::XOnlyPublicKey::new());
```
ssantos21, line 257:
```suggestion
    ) -> Result<PublicKey, MusigTweakErr> {
        let cx = secp.ctx().as_ptr();
        unsafe {
            let mut out = PublicKey::from(ffi::PublicKey::new());
```

ssantos21 (changes requested):
The functions `pubkey_xonly_tweak_add` and `secp256k1_musig_pubkey_xonly_tweak_add` have the wrong parameter (`XOnlyPublicKey` instead of `PublicKey`).

This causes an error when verifying a signature of a tweaked key. Changing them to use the expected parameter fixes this.

src/zkp/musig.rs
```diff
@@ -0,0 +1,1475 @@
...
+    ///
+    /// Different orders of `pubkeys` result in different `agg_pk`s.
+    /// The pubkeys can be sorted lexicographically before combining with which
+    /// ensures the same resulting `agg_pk` for the same multiset of pubkeys.
+    /// This is useful to do before aggregating pubkeys, such that the order of pubkeys
+    /// does not affect the combined public key.
+    ///
+    /// # Returns
+    ///
+    ///  A pair ([`MusigKeyAggCache`], [`XOnlyPublicKey`]) where the first element is the `key_agg_cache`.
```
stevenroose, line 102: This is no longer true.

secp256k1-zkp-sys/src/zkp.rs
sanket1729: Good catch.

sanket1729:
Hello everyone. Sorry for the delay in working on this. 

@ssantos21, there was a good security reason to disallow serialization APIs for SecNonce. I have prefix `dangerous` to those in the name and doc-comment explaining the reason. I am yet sure the use-case of serializing sessions, but we can do that as a follow-up PR. 

I suggest we review this PR for correctness, any incomplete serialization features can be added as a followup. 

Ready for review. cc @stevenroose @jonasnick @Ademan @ssantos21 @apoelstra

src/zkp/musig.rs
```diff
@@ -0,0 +1,1518 @@
...
+    /// This function outputs a secret nonce that will be required for signing and a
+    /// corresponding public nonce that is intended to be sent to other signers.
+    ///
+    /// MuSig differs from regular Schnorr signing in that implementers _must_ take
+    /// special care to not reuse a nonce. If you cannot provide a `sec_key`, `session_id`
+    /// UNIFORMLY RANDOM AND KEPT SECRET (even from other signers).
+    /// Refer to libsecp256k1-zkp documentation for additional considerations.
+    ///
+    /// Musig2 nonces can be precomputed without knowing the aggregate public key, the message to sign.
+    /// See the `new_nonce_pair` method that allows generating [`MusigSecNonce`] and [`MusigPubNonce`]
```
philipr-za, line 304: The `new_nonce_pair` method doesn't seem to exist?

src/zkp/musig.rs
```diff
@@ -0,0 +1,1518 @@
...
+    /// # use secp256k1_zkp::{MusigKeyAggCache, Secp256k1, SecretKey, Keypair, PublicKey, MusigSessionId, Message};
+    /// # let secp = Secp256k1::new();
+    /// # let sk1 = SecretKey::new(&mut thread_rng());
+    /// # let pub_key1 = PublicKey::from_secret_key(&secp, &sk1);
+    /// # let sk2 = SecretKey::new(&mut thread_rng());
+    /// # let pub_key2 = PublicKey::from_secret_key(&secp, &sk2);
+    /// #
+    /// let key_agg_cache = MusigKeyAggCache::new(&secp, &[pub_key1, pub_key2]);
+    /// // The session id must be sampled at random. Read documentation for more details.
+    /// let session_id = MusigSessionId::new(&mut thread_rng());
```
stevenroose, line 340: This method doesn't exist anymore and seems to be renamed to `assume_unique_per_nonce_gen`.

src/zkp/musig.rs
```diff
@@ -0,0 +1,1518 @@
...
+    /// // Provide the current time for mis-use resistance
+    /// let extra_rand : Option<[u8; 32]> = None;
+    /// let (_sec_nonce, _pub_nonce) = key_agg_cache.nonce_gen(&secp, session_id, pub_key1, msg, extra_rand)
+    ///     .expect("non zero session id");
+    /// # }
+    /// ```
+    pub fn nonce_gen<C: Signing>(
+        &self,
+        secp: &Secp256k1<C>,
+        session_id: MusigSessionId,
```
stevenroose, line 353: `MusigSessionId` is not `Copy` and it has to be kept around for partial signature generation, right? So this API should take a ref, I suppose.

stevenroose: Hmm, I seem to be mistaken. The session ID is not needed to create a MusigSession at all. So probably ignore this, but maybe double check :)

stevenroose:
Are you planning serde serialization for the types?

src/zkp/musig.rs
philipr-za: ahh I see it's actually called `new_musig_nonce_pair` now so just a doc-comment issue then

stevenroose:
Suggestion: change the `MusigKeyAggCache` pubkey argument to an iterator so that the user doesn't necessarily have to allocate.

```rust
    pub fn new<C: Verification>(secp: &Secp256k1<C>, pubkeys: &[PublicKey]) -> Self {
```

jlest01:
Is this PR still active? What can be done for the merger?

src/zkp/musig.rs
```diff
@@ -0,0 +1,1518 @@
...
+    ///
+    /// This function outputs a secret nonce that will be required for signing and a
+    /// corresponding public nonce that is intended to be sent to other signers.
+    ///
+    /// MuSig differs from regular Schnorr signing in that implementers _must_ take
+    /// special care to not reuse a nonce. If you cannot provide a `sec_key`, `session_id`
+    /// UNIFORMLY RANDOM AND KEPT SECRET (even from other signers).
+    /// Refer to libsecp256k1-zkp documentation for additional considerations.
+    ///
+    /// Musig2 nonces can be precomputed without knowing the aggregate public key, the message to sign.
```
jonasnick, line 303: Musig2 -> MuSig2, also below in the error enum doc. And sometimes it's just called MuSig. Should be MuSig2 consistently.

jonasnick (commented):
I went throught the PR and marked all comments resolved that appeared to be resolved. There are still a few unaddressed review comments  (e.g., a panic, an incorrect type in the FFI, unfinished sentences). I think the PR also needs to be rebased to use the new v0_10_0 symbols.

@stevenroose

> This makes me wonder, the name "SessionId" somehow made me thing it had to be shared between everyone in the same signing session. But from that sentence maybe it seems that it should be locally random and not shared?

Argh, no, sorry. `SessionId` is absotlutely not to be shared. It must be kept secret from other signers. We renamed this `session_secrand` in the latest version of the module (PR'd to libsecp256k1, not in libsecp256k1-zkp yet).

stevenroose:
Damn we missing @sanket1729 here..

Ademan:
Given https://github.com/bitcoin-core/secp256k1/pull/1479 was merged, does it make sense to focus on https://github.com/rust-bitcoin/rust-secp256k1/pull/716 ?

sanket1729:
Closing this in favor of upstream PR linked in the comment above

PR #49, opened 2022-04-07
[backport -sys 0.4] Fix sys duplicate impl + release -sys v0.4.1

stevenroose:
The -sys v0.4.0 version currently doesn't build. This backports the fix into a new 0.4.1 version. I created a release branch for that.

### DIFF
--- a/secp256k1-zkp-sys/CHANGELOG.md
+++ b/secp256k1-zkp-sys/CHANGELOG.md
@@ -1,3 +1,8 @@
+
+# 0.4.1 - 2022-04-07
+
+- backport compile error fix from 0.5.0
+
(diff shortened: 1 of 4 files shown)


stevenroose:
I've seen some CI fixes by @sanket1729. Which ones do you reckon we need here?

stevenroose:
K I don't really need this anymore, so feel free to close it and you guys enjoy the conference :)

thomaseizinger:
I don't think we can fix this without updating the underlying C lib which we already did in the other PR!

We are also waiting for the downstream ecosystem to upgrade (all the way to bdk and miniscript), so we can use the newly released version 😁

PR #53, opened 2022-07-20
remove redundant hash impls; bump version to 0.5.1

### LINKED ISSUE #52: Cannot build latest tag `0.5.0`
This fixes if we want to release 0.5.1

(FYI the error is also on the other tagged version but the `0.1.0`)

```
diff --git a/Cargo.toml b/Cargo.toml
index 47b2b6f..d9b0d66 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,7 @@ use-rand = ["rand", "secp256k1/rand"]
 
 [dependencies]
 secp256k1 = "0.20.0"
+secp256k1-sys = "=0.4.1"
 secp256k1-zkp-sys = { version = "0.5.0", default-features = false, path = "./secp256k1-zkp-sys" }
 rand = { version = "0.6", default-features = false, optional = true }
 serde = { version = "1.0", default-features = false, optional = true }
```

apoelstra:
Fixes #52

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "secp256k1-zkp"
-version = "0.5.0"
+version = "0.5.1"
 authors = [ "Dawid Ciężarkiewicz <dpc@ucore.info>",
             "Andrew Poelstra <apoelstra@wpsoftware.net>",
             "Lucas Soriano <lucas@comit.network>",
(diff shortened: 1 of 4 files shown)


apoelstra:
Uh, ok, it looks like 0.5.0 is pretty badly broken independently of this..

RCasatta:
Don't bother to fix it if it's a mess, I'll move to `0.6.0` as soon as I understand how to fix #52. Thanks

PR #54, opened 2022-07-25
Backport rust-secp 284 and 422

### LINKED ISSUE #47: Wasm build fails
See: https://github.com/rust-bitcoin/rust-secp256k1/issues/419 for more context

apoelstra:
Fixes #47

(Actually I believe 47 is already fixed, but this re-enables the test in CI so we can be sure.)

Easiest way to review this is probably to add rust-secp as a remote and then use `range-diff` to see how my versions of the commits change from those in the PR. I have provided commit IDs in my commit messages.

### DIFF
--- a/contrib/test.sh
+++ b/contrib/test.sh
@@ -79,3 +79,5 @@ then
         cargo fmt --all -- --check
     )
 fi
+
+exit 0
--- a/contrib/test.sh
+++ b/contrib/test.sh
@@ -59,6 +59,7 @@ fi
 
 # Address Sanitizer
 if [ "$DO_ASAN" = true ]; then
+    clang --version
     cargo clean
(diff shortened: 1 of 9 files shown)


apoelstra:
cc @sanket1729 let's get this in

contrib/test.sh
```diff
@@ -59,6 +59,7 @@ fi
 
 # Address Sanitizer
 if [ "$DO_ASAN" = true ]; then
+    clang --version
```
sanket1729, line 62: There is also another diff here at lin 55:
```diff
 if [ "$DO_WASM" = true ]; then
-    clang --version
+    clang-9 --version
```

sanket1729 (approved):
ACK b0a1a693bb6c4c83ae3e8bd974afa5d66d0217e5. Ignore the comment about clang-9, looks like it is not need for this secp-zkp CI

PR #55, opened 2022-08-31
Fix *Proof::from_str

LeoComandini:
Deserializing `RangeProof`s and `SurjectionProof`s from `str` fails with current master.

Add a fix and tests.


RCasatta:
ACK e3a63b3f68f6ce1deac6f7c8e935e817a176fd20

looks like `from_hex` is an easy-to-call-wrong function

apoelstra:
utACK. Agreed! I cannot run my local tests until #56 is fixed, but I think we're good to merge this.

sanket1729 (approved):
ACK e3a63b3f68f6ce1deac6f7c8e935e817a176fd20

PR #56, opened 2022-08-31
ffi: fix signature of whitelist_sign

apoelstra:
The signature of whitelist_sign in our FFI bindings did not match the signature in the C libarry.

Noticed when running local tests on #55. Apparently wasm-pack notices stuff like this now and errors out :)


jonasnick (commented):
utACK 98994eed5d3b488b774521226c6f345f477ab7d5

apoelstra:
Heads up this will require a major release of both secp-zkp and secp-zkp-sys.

PR #58, opened 2022-09-25
Update rust-secp256k1 and rust-secp256k1-sys

justinmoon:
Closes #57 

I removed the `unstable` feature because [it was removed from rust-secp256k1](https://github.com/rust-bitcoin/rust-secp256k1/commit/71b47d127396e1e246fb06fd86b2ad3b920304ee). Let me know if we should keep this.

`secp256k1::secp256k1_sys` re-export was removed, but it seems that `ffi` re-export is the same thing?

PR #59, opened 2022-09-26
secp256k1: bump 0.22.1 -> 0.24.0

RCasatta:
supersede #58


RCasatta:
Commits need squashing/reorgs but it's ready for review

apoelstra:
Done reviewing aa2a42525ea1ed9f8f5cbf15b3114f1d2d169958.

Note that updating `rand` breaks our ability to combine for wasm when the `rand-std` feature is enabled. (Though the old version of `rand` probably does something insane, so.. see https://github.com/rust-bitcoin/rust-secp256k1/issues/346 for additional context/discussion.)

Currently the individual commits don't compile. I'd suggest reordering/squashing to:

* Bump MSRV in github config; set edition to 2018; do edition fixes (including those when `rand` is enabled)
* Bump secp and rand dependency versions
* Change to avoid deprecated functions
* cargo fmt

RCasatta:
I reorganized the commit as suggested.

I think CI must be re-kicked because errors are due to network issues while downloading from crates.io

> Note that updating rand breaks our ability to combine for wasm when the rand-std feature is enabled. (Though the old version of rand probably does something insane, so.. see https://github.com/rust-bitcoin/rust-secp256k1/issues/346 for additional context/discussion.)

After reading I am still not sure how to handle this, should I use `getrandom` for the wasm target like it's done in `rust-secp256k1`?

apoelstra:
> After reading I am still not sure how to handle this, should I use `getrandom` for the wasm target like it's done in `rust-secp256k1`?

I think what you've done (basically ignore the situation) is fine. Users who want to use wasm will need to compile with `no-std` and they'll lose some functionality.

In general we have trouble with wasm right now for rust-secp and derivatives. See https://github.com/rust-bitcoin/rust-secp256k1/issues/485. I'd like to explore reimplementing some parts of libsecp (at least parsing, and maybe verification) directly in Rust, but that's a long-term project.

apoelstra (approved):
ACK 850e5de7407b6271eaa0eb789e7774f2bc1539ff

sanket1729 (approved):
ACK 850e5de7407b6271eaa0eb789e7774f2bc1539ff

PR #62, opened 2022-12-26
Update to secp 0.26 and secp-sys 0.8

Tibo-lg:
This PR updates the rust-secp256k1(-sys) dependencies to latest released ones. I tried to mimic what was done upstream. Regarding the trait implementations (`Eq`, `PartialEq`,...), since the sys types in this crate don't seem to have `serialize` functions I felt like it was ok to just derive then but please correct me if that's wrong.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -74,11 +74,9 @@ jobs:
           - 1.41.1
           - beta
           - stable
-        target: [ x86_64-unknown-linux-gnu, x86_64-apple-darwin ]
-        include:
-          - target: x86_64-unknown-linux-gnu
-            os: ubuntu-latest
-          - target: x86_64-apple-darwin
+        os: [ ubuntu-latest, macos-latest ]
+        exclude:
+          - rust: 1.41.1
             os: macos-latest
     runs-on: ${{ matrix.os }}
     steps:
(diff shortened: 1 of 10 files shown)


Tibo-lg:
I had to implement the traits to make it work with 1.41.1 but not sure what is causing the CI failures. I'll have a look at it again but if someone has a suggestion let me know.

Tibo-lg:
I'm quite confused, unless I'm missing something it looks like actions on the same commit passed in my forked repo:
https://github.com/p2pderivatives/rust-secp256k1-zkp/actions/runs/3781248443

Tibo-lg:
Investigating the CI failure I found that for some reason it only happens on recent version of OSX (I've updated recently and before the update I could not reproduce the issue, now I can). The issue also seem to happen on master and also on `rust-secp256k1`, so I'm pretty confident it's not introduced by this PR.

apoelstra (approved):
ACK fa5d316d4597e09221b34daa2ab44e29916c6377

apoelstra:
CI failure seems to be same as https://github.com/rust-pcap/pcap/issues/280 (to take an example from a random project). Apparently "rust 1.41 does not work on macos >= 12". So we will just need to adjust CI.

@Tibo-lg could you add a commit which changes CI to not test 1.41 on Mac?

Tibo-lg:
> @Tibo-lg could you add a commit which changes CI to not test 1.41 on Mac?

@apoelstra Done! I've modified a bit the git action file removing the target field which seemed unused, making it easier to exclude 1.41 on macos, but I hope that doesn't break anything.
Thanks for checking the PR!

apoelstra:
Looks good. I'll ACK and merge this since it fixes CI.

But cc @jonasnick is there any specific reason we had the target field here? Was it to test different endianness or anything?

apoelstra (approved):
ACK a1cff98794ea9bc7eca77aba771db54de524a706

jonasnick:
I don't remember. It seems redundant.

PR #63, opened 2023-04-01
Update to upstream 1d256089004a19bdbead7c5676e52c8e07b09fce

sanket1729:
Can re-run by `./vendor_libsecp.sh` from ./contrib


apoelstra:
Do you think we could copy the new revendor.sh from libsecp over here first? It would let me review this PR by completely ignoring the contents/changes of `depend/` and just checking that the libsecp commit is something sane.

sanket1729:
Superceded by #64

PR #64, opened 2023-04-13
Replace revendoring script; update to upstream 1d256089004a19bdbead7c5676e52c8e07b09fce

apoelstra:
The first commit copies the new revendoring script from rust-secp; it can be verified by just diffing the two files. There are minimal changes.

The other two commits are copies of #63 which can be verified using `git range-diff`. (The commit which updates upstream was actually created fresh using the new revendoring script; but the result is identical, as expected.)


sanket1729 (approved):
ACK 8e2d798e129c463700bd9ae73dd75aaf5d584dc1

sanket1729:
Also reviewed that the new vendoring script is same as that of secp upstream.

PR #65, opened 2023-05-08
nits from `cargo test` and `cargo clippy` on latest nightly

apoelstra:
The newest nightly has a new "unused mut" warning which is causing CI to fail. Fix that, and while we're at it, run clippy. And while we're at that, remove some `#[deny]` conditions from the crate root since these properly belong in CI (via -D warnings) rather than in the source.

### DIFF
--- a/secp256k1-zkp-sys/src/zkp.rs
+++ b/secp256k1-zkp-sys/src/zkp.rs
@@ -426,6 +426,12 @@ impl PartialEq for SurjectionProof {
 
 impl Eq for SurjectionProof {}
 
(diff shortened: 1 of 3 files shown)


sanket1729:
LGTM, will ACK after #54 is in.

apoelstra:
Ok, rebased and should be good to go. This redoes a bit of the "add doccomments to re-exports" word but I think it does a better job, so I kept the new version.

sanket1729 (approved):
ACK e016ba9d13493f7ad84430ce040133485d7e7941

PR #69, opened 2023-07-14
conditionally derive Ord

RCasatta:
Otherwise in fuzzing configuration there is a double impl

Noticed the error in rust-elements CI

### DIFF
--- a/secp256k1-zkp-sys/src/zkp.rs
+++ b/secp256k1-zkp-sys/src/zkp.rs
@@ -533,7 +533,8 @@ impl From<Tag> for [u8; 32] {
 
 // TODO: Replace this with ffi::PublicKey?
 #[repr(C)]
-#[derive(Copy, Clone, PartialOrd, Ord)]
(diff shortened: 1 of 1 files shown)


apoelstra:
If you `git grep fuzzing` the only things that appear are `not(fuzzing)`. I'm confused as to where the double-impl would be coming from.

RCasatta:
the `cfg(fuzzing)` is coming from the `impl_array_newtype!` macro in secp256k1-sys

the error is in rust-elements fuzzing job

https://github.com/ElementsProject/rust-elements/actions/runs/5554974806/jobs/10145494889

apoelstra:
Heh, ok, good to know. That's a bit fragile but I guess it's fine for now. Thanks for noticing!

apoelstra:
Later versions of rust-secp rename `fuzzing` to `rust_secp_fuzzing` or something. We'll have to decide here whether to use the same name or to add a new `rust_secp_zkp_fuzzing`.

PR #76, opened 2024-01-03
secp256k1-zkp-sys: remove not(fuzzing) gates on derives

apoelstra:
I'm not sure why these derives are gated on not(fuzzing); I believe that there used to be fuzzing-only manual implementations. These do not exist anymore (or maybe they only existed upstream), and as a result the library does not compile when fuzzing is enabled.

Also, make sure that cfg(fuzzing) is replaced by cfg(rust_secp_fuzz) everywhere so that the user has the ability to disable fuzzing mode.

### DIFF
--- a/secp256k1-zkp-sys/CHANGELOG.md
+++ b/secp256k1-zkp-sys/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 0.9.1 - 2023-01-03
+
+- Many changes; restart CHANGELOG.
+
 # 0.2.0 - 2021-01-06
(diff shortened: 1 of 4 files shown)


RCasatta:
ACK ee43b6147becf3db4b77fa8e95d4c8187d1122e9

apoelstra:
Published. I did not make a tag for it since it appears that secp-zkp-sys does not have any tags currently, and this commit is easy to locate via `git blame`. But maybe we should change that.

PR #80, opened 2024-06-10
Update secp256k1 dependency

### LINKED ISSUE #79: Update `secp256k1` to `0.29`
`bitcoin@0.32` uses `secp256k1@0.29`

apoelstra:
Also do some other deferred maintenance.

Fixes #79.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -107,4 +107,4 @@ jobs:
           toolchain: ${{ matrix.rust }}
           override: true
       - name: Running cargo in release mode
-        run: cargo test --features="global-context" --release
+        run: cargo update -p cc --precise 1.0.94 && cargo test --features="global-context" --release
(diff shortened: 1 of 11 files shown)


bennyhodl:
Is this going to be merged and released soon?

jonasnick (approved):
ACK 4fb4de38f8e412a7acdaf10160b9f36cb3a5aa39

apoelstra:
> Is this going to be merged and released soon?

I'd like to update the copy of libsecp256k1-zkp as well to fix https://github.com/ElementsProject/rust-elements/issues/202 and then yes, an update is a high priority.

@jonasnick do you know what the relationship between libsecp and libsecp-zkp is right now? Can I just update the libsecp-zkp in this library to `master` and expect things to work? I believe we depend here on the context objects from both libraries to be identical.

jonasnick:
Yes I'd expect that updating libsecp-zkp in this library will work. Good point about the context object; it's identical in both forks of the library.

PR #81, opened 2024-07-09
Update vendored version of secp256k1-zkp

apoelstra:
Also bump major version of both secp256k1-zkp and secp256k1-zkp-sys.

Next step toward https://github.com/ElementsProject/rust-elements/issues/202#issuecomment-2199980614


jonasnick (commented):
I think this misses a change log entry, other than that ACK.

apoelstra:
Added a commit with CHANGELOG entries.

jonasnick (approved):
ACK c1feb16e960058bc3771da2d186d90b4e8773766

apoelstra:
Tagged, and published both crates.

PR #86, opened 2024-10-18
fix: WASM

uncomputable:
Port of https://github.com/rust-bitcoin/rust-secp256k1/pull/735

Let's see if this passes CI. I can run `cargo build --target wasm32-unknown-unknown` after my commit.

Generating the `.patch` files was kind of hard. Is there an easy way?


uncomputable:
There are strictly more checkmarks than on master :)

apoelstra (approved):
ACK 454a9aa7119624300eab19e93fb5710c29b0881a; successfully ran local tests; thanks!!

PR #89, opened 2025-04-06
Add Bindings for Schnorr Adaptor Signatures

siv2r:
This PR introduces Rust bindings for the Schnorr adaptor signature functionality added in: https://github.com/BlockstreamResearch/secp256k1-zkp/pull/299. The secp256k1-zkp PR is pending review from the maintainers, but I don’t expect the user-facing APIs to change much. Hence, I’m opening the bindings PR here.

**Changes:**
- Provides FFI interfaces to the corresponding libsecp256k1-zkp functions:
    - `secp256k1_schnorr_adaptor_presign`
    - `secp256k1_schnorr_adaptor_extract`
    - `secp256k1_schnorr_adaptor_adapt`
    - `secp256k1_schnorr_adaptor_extract_sec`
- Introduces a dedicated module for adaptor signature bindings:
    - adds the `zkp::schnorr_adaptor::SchnorrAdaptorPreSignature` wrapper type.
    - Implements methods to:
        - Create pre-signatures with `presign`.
        - Extract the adaptor point via `extract_adaptor`.
        - Adapt a pre-signature to a valid Schnorr signature using `adapt`.
        - Extract the secret adaptor through `extract_secadaptor`.
    - Includes a correctness test to ensure proper functionality.

**Note:** The overall diff might appear large because the first commit (528f610) updates the secp256k1-zkp-sys vendor with the Schnorr adaptor module (since those changes haven’t been merged upstream yet). The code changes relevant to the bindings are in commits 2 and 3.


apoelstra:
Nice, thanks!! I reviewed the code, though I don't think we should merge it until upstream merges theirs and we can vendor it from the main repo.

src/zkp/schnorr_adaptor.rs
```diff
@@ -0,0 +1,298 @@
...
+
+impl fmt::Display for SchnorrAdaptorPreSignature {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        fmt::LowerHex::fmt(self, f)
+    }
+}
+
+impl str::FromStr for SchnorrAdaptorPreSignature {
+    type Err = Error;
+    fn from_str(s: &str) -> Result<SchnorrAdaptorPreSignature, Error> {
```
apoelstra, line 36: In 28fd225cc653682abc5b320f7c45c0ca2d0b7b36:

Can you use `Self` rather than `SchnorrAdaptorPreSignature` everywhere you can? It helps to keep track of which functions are manipulating an object and which ones are converting it to other types.

src/zkp/schnorr_adaptor.rs
```diff
@@ -0,0 +1,298 @@
...
+        SchnorrAdaptorPreSignature::presign_with_aux_rand(secp, msg, keypair, adaptor, &aux)
+    }
+
+    /// Creates a Schnorr adaptor pre-signature without using any auxiliary
+    /// random data. Note that using this function is still considered safe.
+    pub fn presign_no_aux_rand<C: Signing>(
+        secp: &Secp256k1<C>,
+        msg: &Message,
+        keypair: &Keypair,
+        adaptor: &PublicKey,
```
apoelstra, line 161: In 28fd225cc653682abc5b320f7c45c0ca2d0b7b36:

I think we should introduce a new type `PublicAdaptor`. Fine to have `From<PublicKey>` and vice-versa on it, but it helps to keep the types separate. Also fine, I think, to keep using `SecretKey` for the secret adaptor since that's only used in one function and it's one you've gotta be careful about anyway.

apoelstra: Oh, actually I see that we have methods that both create a signature from a secret adaptor *and* extract it. Now I think that we should create a new `SecretAdaptor` (or maybe just `Adaptor`) type for the secret half as well.

src/zkp/schnorr_adaptor.rs
```diff
@@ -0,0 +1,298 @@
...
+                aux_rand.as_c_ptr() as *const ffi::types::c_uchar,
+            )
+        };
+        debug_assert_eq!(res, 1);
+
+        SchnorrAdaptorPreSignature(pre_sig)
+    }
+
+    /// Extracts the adaptor point from a Schnorr adaptor pre-signature.
+    pub fn extract_adaptor(&self, msg: &Message, pubkey: &XOnlyPublicKey) -> Result<PublicKey, Error> {
```
apoelstra, line 207: In 28fd225cc653682abc5b320f7c45c0ca2d0b7b36:

In this method, `extract_secret_adaptor`, and `adapt`, I'm tempted to take `self` by value so that the user can't use the same presignature with multiple adaptors or signatures unless they call `.clone` to explicitly duplicate it. This is a mild inconvenience, but the `clone` call is an opportunity for the user to comment what it is they're doing using the same (pre)sig in two different contexts, and for simple protocols it might prevent accidents.

On the other hand, this might be seen as just "making the API hard to use for no good reason" since a presignature is a public object and there's no actual harm in trying to do multiple things with it. (Plus, this entire adaptor signature module is designed for "roll your own crypto" applications.)

@jonasnick what do you think?

src/zkp/schnorr_adaptor.rs
siv2r: Okay, I'll introduce the `SecretAdaptor` and `Adaptor` types. The `Adaptor` type for the public adaptor, to keep it consistent with the zkp module naming convention.

src/zkp/schnorr_adaptor.rs
jonasnick: I don't really have an opinion on this. It seems like given that the module is designed to "roll your own crypto", the accidents that can happen by using the same presig in two contexts seem to be not very severe in comparison.

apoelstra: Yeah, that's how I'm leaning. Ok, let's *not* take `self` by value.

PR #93, opened 2026-04-14
Eliminate dependency on the rand crate

### LINKED ISSUE #92: rand crate dependency triggers security warning
(moved from my erroneous filing at https://github.com/BlockstreamResearch/secp256k1-zkp/issues/358 )

Per https://rustsec.org/advisories/RUSTSEC-2026-0097 , rand 0.8.* has a vulnerability. I don't think secp256k1-zkp is affected directly, but its reference of that version of rand causes security scanners (e.g., cargo audit -D warnings) to indicate a vulnerability.

@apoelstra commented at https://github.com/BlockstreamResearch/secp256k1-zkp/issues/358#issuecomment-4240498734 : 

```
Yeah let's just delete this dep. It has caused us nothing but suffering.

If users want to generate a random key, they can use a separate getrandom (or whatever) dependency to get 32 random bytes and then use SecretKey::from_secret_bytes.
```

jhfrontz:
Closes #92


apoelstra:
utACK dd0ff0fadc1a00129271d6a7de90aa0e6a2ce415

delta1:
CI will need to be fixed since it is specifying a `rand` feature 

> error: package `secp256k1-zkp v0.11.0 (/home/runner/work/rust-secp256k1-zkp/rust-secp256k1-zkp)` does not have the feature `rand`

delta1:
Lines 3, 38, 39 in https://github.com/BlockstreamResearch/rust-secp256k1-zkp/blob/master/contrib/test.sh

jhfrontz:
> Lines 3, 38, 39 in https://github.com/BlockstreamResearch/rust-secp256k1-zkp/blob/master/contrib/test.sh

Oh, my -- I clearly don't understand github CI (I was wondering why `cargo test` exited with no tests having run).

delta1:
Running contrib/test.sh with some of these env vars set should replicate it 

https://github.com/BlockstreamResearch/rust-secp256k1-zkp/blob/master/.github/workflows/rust.yml#L59

PR #94, opened 2026-04-14
Bump rand version to 0.9

### LINKED ISSUE #92: rand crate dependency triggers security warning

jhfrontz:
Closes #92 

(alternate to #93 )

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ rand = ["actual-rand", "secp256k1/rand"]
 
 [dependencies]
 actual-serde = { package = "serde", version = "1.0", default-features = false, optional = true }
-actual-rand = { package = "rand", version = "0.8", default-features = false, optional = true }
+actual-rand = { package = "rand", version = "0.9", default-features = false, optional = true }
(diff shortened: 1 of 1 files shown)


apoelstra:
Closing in favor of #93. rand 0.9 does not support our MSRV.

(I think CI is failing for unrelated reasons; we really need to update this crate..)

PR #95, opened 2026-04-14
Eliminate external dependency on the rand crate

jhfrontz:
Alternate solution to #92

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,23 +22,22 @@ rustdoc-args = ["--cfg", "docsrs"]
 [features]
 default = ["std"]
 std = ["secp256k1-zkp-sys/std", "secp256k1/std"]
-rand-std = ["actual-rand/std", "secp256k1/rand-std"]
+rand-std = ["secp256k1/rand-std"]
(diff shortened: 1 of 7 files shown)


apoelstra:
Yeah, this seems like a good stopgap. We should rewrite these tests eventually though. Upstream I believe we replaced them all (or at least, most of them) with ones that use a simple deterministic RNG.

jhfrontz:
I fixed the `fmt` but I don't understand the other two CI failures.

apoelstra:
I suspect the other two failures are true on `master`. This looks severely bitrotted. We have not merged any PRs since October of 2024.

jhfrontz:
@apoelstra I wondered about that -- the one failure is about the libc version (which seems beyond the change proposed here).  I'm not sure how to proceed?

apoelstra:
I'd be happy to just merge this -- but I want to fix the CI problems before we can cut another release, which presumably you want so you can push the `rand` dependency out of your stack.

jhfrontz:
Right, bigger things in the pipeline -- `rand` maintainer looks to be trying to backport the fix to 0.8.x, so maybe that's our stopgap.

PR #96, opened 2026-06-18
chore: sync with rust-secp256k1 0.31.1

mllwchrry:
The sync consists of three commits:

- **Commit A**: bump `secp256k1-zkp` to 0.12.0 and `secp256k1-zkp-sys` to 0.11.0, update `vendor-libsecp.sh` (new repo URL, updated patches), re-vendor the C library with `rustsecp256k1zkp_v0_11_0_` symbols.
- **Commit B**: copy `src/` and `secp256k1-zkp-sys/src/` from `rust-secp256k1` 0.31.1, rename symbols, remove `secp256k1-sys` dep from sys crate.
- **Commit C**: reconnect `src/zkp/`, add ZKP error variants, fix rand 0.9 API, upgrade rand dep to 0.9, add `ENABLE_MODULE_ECDH` / `ENABLE_MODULE_RECOVERY` to `build.rs`.

Follow-up PR will make `secp256k1` an optional dep (on by default) and then delete the locally-copied ordinary signing code in favour of `pub use secp256k1::*`.


apoelstra:
Let me review and test this locally. I'm not sure that we expect our CI to pass yet. Certainly not the WASM tests.

mllwchrry:
Fixed several CI issues after the initial push:

- Renamed `hashes` dep to `actual-hashes` (the `dep:` prefix syntax requires Cargo 1.60+).
- Added `rand` to the `rand-std` feature dependencies so `crate::rand` is available when `rand-std` is enabled without `rand`.
- Fixed `use rand::` to be `use crate::rand::` in all src/zkp/ files (broken by the `actual-rand` dep rename).
- Removed the `no_std` attribute from src/lib.rs because it caused `cargo build --no-default-features` to fail since `cdylib` requires a panic handler in `no_std` mode. The old codebase didn't have `no_std` support either, so this restores previous behavior.
- Bumped CI MSRV from 1.56.1 to 1.65.0 (secp256k1 0.31.1 states 1.63.0, but its transitive dep `libc` requires 1.65), removed the cargo update -p cc --precise 1.0.94 workaround.

The only remaining CI failures are the pre-existing WASM issues.

apoelstra:
In 36daa02eb1408b33264b057575662ef61c5862d4:

Can you redo this commit without all the formatting changes?

apoelstra:
In 21bd9a329b544bcc2a6fcff73e0eebda04952aa2:

This also seems to have a ton of formatting differences vs rust-secp.

mllwchrry:
The formatting differences in commit 21bd9a3 were probably coming from commit 36daa02 being shown cumulatively by GitHub. Commit 21bd9a3 itself only has the necessary symbol renames vs upstream rust-secp. I've now separated the formatting into its own commit 794407e for the tests to pass. It should be much easier to review now.

apoelstra:
> The formatting differences in commit [21bd9a3](https://github.com/BlockstreamResearch/rust-secp256k1-zkp/commit/21bd9a329b544bcc2a6fcff73e0eebda04952aa2) were probably coming from commit [36daa02](https://github.com/BlockstreamResearch/rust-secp256k1-zkp/commit/36daa02eb1408b33264b057575662ef61c5862d4) being shown cumulatively by GitHub

I don't ever use Github to view diffs. It's not a trustworthy interface.

I'm running `diff -r . ~/code/rust-bitcoin/rust-secp256k1/master/secp256k1-sys` on the commit and I see tons of stuff like

```
1215,1217c1649,1654
<     pub unsafe fn secp256k1_ec_pubkey_parse(cx: *const Context, pk: *mut PublicKey,
<                                             input: *const c_uchar, in_len: size_t)
<                                             -> c_int {
---
>     pub unsafe fn secp256k1_ec_pubkey_parse(
>         cx: *const Context,
>         pk: *mut PublicKey,
>         input: *const c_uchar,
>         in_len: size_t,
>     ) -> c_int {
```

What commit are you pulling rust-secp256k1 from?

mllwchrry:
I pulled from the secp256k1-0.31.1 tag (commit 0dfa825a). Commit 21bd9a3 is a straight copy of those files with only symbol renames. I used 0.31.1 rather than master because the top-level crate depends on secp256k1 = "0.31.1".

The formatting you see in 21bd9a3 is exactly what 0.31.1 contains, master got cargo fmt applied to secp256k1-sys in f5f90af2 after the 0.31.x branch diverged. Commit 794407e applies cargo fmt on top.

apoelstra:
@mllwchrry thanks! I understand the problem now. I had been diffing aganist rust-secp256k1 master.

21bd9a329b544bcc2a6fcff73e0eebda04952aa2 looks good to me! Thanks for investigating the history of the format divergences.

apoelstra:
Gonna try to run my local CI over all these commits -- but after reviewing the code this looks perfect. Let's remember to redo the "copy from rust-secp256k1 and rename symbols" commit after the next rust-secp release.

apoelstra:
As for the CI failure here, let's fix that in a separate PR.

apoelstra (approved):
ACK 794407eb1f42fd46ea99b2c66384d0f745fdace1; successfully ran local tests

apoelstra:
The ACK message above is a bit misleading. I had to skip a0ce04049fec1225f6761afa821e5b63935f87c5 and 21bd9a329b544bcc2a6fcff73e0eebda04952aa2 which don't build; I had to disable clippy; I had to disable the doc-tests; I had to hardcode a lockfile and a MSRV. But I got things working :)

apoelstra:
This PR contained version bumps to secp256k1-zkp and secp256k1-zkp-sys but I did **not** tag or publish them. We will have to do that manually once the library is in a better state.

PR #97, opened 2026-07-12
A bunch of cleanups to get my local CI system working

apoelstra:
This is a first pass on cleaning up some of the code and fixing lints, etc. This should get my local CI system to pass (and maybe even Github CI).

I did my best to avoid touching any code copied from rust-secp256k1, since we intend to delete that code in favor of re-exports and conversion functions.


apoelstra:
Will need to update my lockfile since not all crates will compile with 1.74..

apoelstra:
On 509f399a343f8d733e2e14051c99ba443d16d2a1 successfully ran local tests

src/zkp/ecdsa_adaptor.rs
```diff
@@ -129,9 +130,9 @@ impl EcdsaAdaptorSignature {
     /// Creates an adaptor signature along with a proof to verify the adaptor signature.
     /// This function derives a nonce using a similar process as described in BIP-340.
     /// The nonce derivation process is strengthened against side channel
-    /// attacks by providing auxiliary randomness using the ThreadRng random number generator.
-    /// Requires compilation with "rand-std" feature.
-    #[cfg(feature = "rand-std")]
+    /// attacks by providing auxiliary randomness using the [`rand::thread_rng`] random number generator.
```
mllwchrry:
```suggestion
    /// attacks by providing auxiliary randomness using the [`rand::rng`] random number generator.
```

src/zkp/tag.rs
```diff
@@ -39,8 +40,10 @@ impl fmt::LowerHex for Tag {
     }
 }
 
-#[cfg(all(test, feature = "rand-std"))]
+#[cfg(feature = "std")]
+#[cfg(feature = "rand")]
```
mllwchrry, line 45: It looks like the `#[cfg(test)]` gate was dropped when the `rand-std` feature was removed. Was it intentional?

.github/workflows/rust.yml
```diff
@@ -89,12 +86,12 @@ jobs:
     strategy:
       matrix:
         rust:
-          - 1.65.0
+          - 1.74.0
           - beta
           - stable
         os: [ ubuntu-latest, macos-latest ]
         exclude:
```
mllwchrry: Should we remove this `exclude` the same way it was done for `Tests`?

.github/workflows/rust.yml
mllwchrry: Was this meant to be removed along with the WASM support?

mllwchrry (commented):
Nice cleanup. Left some inline comments, mostly small things.

apoelstra:
Thanks! Fixed all four nits.

apoelstra:
On 59f75955c27ce21b68655597aac9e19d6e0fc75d successfully ran local tests (though skipped a couple of the first commits)

mllwchrry:
ACK 59f7595

PR #98, opened 2026-07-13
replace upstream functionality with re-exports

apoelstra:
This reduces the library back to being "just the zkp module", with the upstream rust-secp256k1 functionality replaced by re-exports from rust-secp256k1.

This requires users juggle two separate context objects, but we haven't found a good solution to that without needing significant changes from upstream. (Both this library and upstream are moving toward an "implicit global context" model, which will hide this duplication from the user, but it'll still be there in the form of multiple redundant in-memory context objects that don't share rerandomization.)

In a later PR we can try deleting the rust bindings for signature stuff from secp256k1-zkp-sys and see if maybe this makes the linker delete a lot of cruft? I'm skeptical since all the core group/field logic will remain.

Anyway, rather than waiting for upstream refactors, let's move forward modernizing this libary and just accept the duplication for now.


apoelstra:
ACK ffffd14c6c77e925a29d9b19451b5a950a442401; successfully ran local tests

src/ffi_convert.rs
```diff
@@ -0,0 +1,86 @@
...
+
+impl ConvertFromUpstream for &secp256k1::ecdsa::Signature {
+    type Target = *const ffi::Signature;
+
+    fn to_zkp_ffi(self) -> Self::Target {
+        use secp256k1::ffi::CPtr as _;
+
+        // SAFETY: technically, both these pointer casts are safe, and it's only downstream
+        // dereferences that require unsafe{} blocks. But such dereferences are guaranteed
+        // to be sound because both libraries have identical representations of ECDSA sigantures
```
mllwchrry:
```suggestion
        // to be sound because both libraries have identical representations of ECDSA signatures
```

src/zkp/whitelist.rs
```diff
@@ -400,55 +424,57 @@ mod tests {
             let decoded = WhitelistSignature::from_slice(&encoded).unwrap();
             assert_eq!(
                 Err(Error::InvalidWhitelistProof),
-                decoded.verify(SECP256K1, &pak_online, &pak_offline, &whitelist_pk,)
+                test_util::whitelist_verify(SECP256K1, &keyset, &decoded),
```
mllwchrry: This should probably use `&original_keyset` since by this point `keyset` has been overwritten with a wrong `whitelist_pk`, so the assertion passes regardless of the tampered bytes.

src/lib.rs
```diff
@@ -623,15 +567,18 @@ mod tests {
...
+        let keyset = whitelist_test_util::KeySet::new(10);
+
         // Try signing
-        assert_eq!(sign.sign_ecdsa(msg, &sk), full.sign_ecdsa(msg, &sk));
-        let sig = full.sign_ecdsa(msg, &sk);
+        assert_eq!(
+            whitelist_test_util::whitelist_prove(&sign, &keyset, 0),
+            whitelist_test_util::whitelist_prove(&full, &keyset, 0),
+        );
+        let sig = whitelist_test_util::whitelist_prove(&sign, &keyset, 0).unwrap();
```
mllwchrry, line 577: It might be negligible, but `whitelist_prove(&sign, &keyset, 0)` runs twice for the same result. `assert_eq!` drops the result, so it's recomputed on the next line. Same pattern in `test_preallocation` and `capabilities`.

mllwchrry (commented):
Great PR! I have not found any serious issues, only two nits and one test logic regression.

src/lib.rs
apoelstra: Yeah, I noticed that when I was writing the code but I decided to leave it since the cost is pretty small and it makes the test a little nicer to read.

apoelstra:
On 4a010056b8e86943b8c81ddef3a698962b463106 successfully ran local tests

mllwchrry:
ACK 4a01005.

PR #99, opened 2026-07-27
Rust bindings for musig2

Sdoba16:
Adds Rust bindings for MuSig2, wrapping the C implementation from
[secp256k1-zkp](https://github.com/BlockstreamResearch/secp256k1-zkp).

Includes:
- FFI bindings in `secp256k1-zkp-sys` (`ENABLE_MODULE_MUSIG`).
- Safe wrappers: key aggregation, nonce generation, partial signing/verification
  and aggregation, plain/x-only tweaking, and adaptor signatures.
- `serialize`/`from_slice`, hex `Display`/`FromStr`, and `serde` impls.
- Unit tests for signing, serialization, tweaking, and adaptor round-trips.

### DIFF
--- a/secp256k1-zkp-sys/build.rs
+++ b/secp256k1-zkp-sys/build.rs
@@ -43,6 +43,7 @@ fn main() {
         .define("ENABLE_MODULE_RANGEPROOF", Some("1"))
         .define("ENABLE_MODULE_ECDSA_ADAPTOR", Some("1"))
         .define("ENABLE_MODULE_WHITELIST", Some("1"))
+        .define("ENABLE_MODULE_MUSIG", Some("1"))
         .define("ECMULT_GEN_PREC_BITS", Some("4"))
         // TODO these three should be changed to use libgmp, at least until secp PR 290 is merged
         .define("USE_NUM_NONE", Some("1"))
--- a/secp256k1-zkp-sys/src/lib.rs
+++ b/secp256k1-zkp-sys/src/lib.rs
@@ -32,6 +32,9 @@ pub mod recovery;
 pub mod zkp;
 pub use zkp::*;
 
+pub mod musig;
+pub use musig::*;
+
 use core::ptr::NonNull;
 use core::{ptr, slice};
 use types::*;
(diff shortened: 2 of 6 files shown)


apoelstra:
How does this compare to https://gitea.bitcoin.ninja/rust-bitcoin/rust-secp256k1/pulls/716 ?

I'd much prefer if we updated the rust-secp dependency to the latest prerelease version which already has musig https://docs.rs/secp256k1/0.32.0-beta.2/secp256k1/musig/index.html

I'll see about just cutting a real release soon.

Sdoba16:
> How does this compare to https://gitea.bitcoin.ninja/rust-bitcoin/rust-secp256k1/pulls/716 ?

I actually did it from scratch, and hadn't seen it, agree it's better to use that.

> I'd much prefer if we updated the rust-secp dependency to the latest prerelease version which already has musig https://docs.rs/secp256k1/0.32.0-beta.2/secp256k1/musig/index.html
> 
> I'll see about just cutting a real release soon.

Sounds great, thanks, appreciate it! I will wait for the release.

apoelstra:
> Sounds great, thanks, appreciate it! I will wait for the release.

Let's say, if there's no release in the next 7 days, we should open a new PR where we just copy/paste the code from rust-secp256k1. I think it `impl`s methods on `PublicKey` and `schnorr:Signature` so we'd turn those into extension traits.

Is that too long? If so, let's just skip straight to "open a PR to copy/paste the code".

Sdoba16:
> Let's say, if there's no release in the next 7 days, we should open a new PR where we just copy/paste the code from rust-secp256k1. I think it `impl`s methods on `PublicKey` and `schnorr:Signature` so we'd turn those into extension traits.
> 
> Is that too long? If so, let's just skip straight to "open a PR to copy/paste the code".

Seven days works for us, no problem. Sounds great!

apoelstra:
Sorry, it's been much more than 7 days, but we did release the new rust-secp256k1 version. I am going to close this and we should have a PR to update the dep here shortly.

PR #101, opened 2026-08-04
 zkp: check input lengths in from_slice before FFI parsing

ethicnology:
`PedersenCommitment::from_slice` and `Generator::from_slice` are safe functions wrapping FFI parsers (`secp256k1_pedersen_commitment_parse`, `secp256k1_generator_parse`) that take no length argument and unconditionally read 33 bytes from the slice pointer. Passing an empty or short slice is therefore an out-of-bounds read; on an empty slice the dangling pointer (address 0x1) is dereferenced and the process segfaults instead of returning an error.

This makes the safe `from_slice` API unsound for any caller parsing untrusted input. A concrete downstream instance: rust-elements deserializes PSET commitment fields (`issuance_value_comm`, `issuance_inflation_keys_comm`, `amount_comm`, `asset_comm`) through these functions, so a malicious PSET segfaults any application parsing it. Found by libFuzzer/AddressSanitizer on the rust-elements `deserialize_pset` target; the rust-elements side is fixed in a separate PR.

Reject any length other than 33 bytes before reaching FFI. The C parsers have no length parameter and read exactly one prefix byte plus a 32-byte x coordinate, so 33 is the only valid input size; the guard restricts no legitimate use. Adds regression tests for empty, short and invalid-content inputs.

The other FFI parse entry points were audited and are not affected: `surjectionproof_parse`, `whitelist_signature_parse` and `rangeproof_info` all validate the input length before dereferencing.

related to https://github.com/BlockstreamResearch/rust-secp256k1-zkp/pull/101

### DIFF
--- a/src/zkp/generator.rs
+++ b/src/zkp/generator.rs
@@ -132,6 +132,12 @@ impl Generator {
 
     /// Parse a generator from a slice of bytes.
     pub fn from_slice(bytes: &[u8]) -> Result<Self, Error> {
+        // The FFI parser takes no length and unconditionally reads 33 bytes
+        // from the slice pointer; reject other lengths here to avoid an
+        // out-of-bounds read (a segfault on empty input).
+        if bytes.len() != 33 {
+            return Err(Error::InvalidGenerator);
+        }
         let mut public_key = unsafe { ffi::PublicKey::new() };
 
         let ret = unsafe {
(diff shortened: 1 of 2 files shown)


apoelstra:
Thanks for beating me to the punch on this!

apoelstra:
I think we should also backport this to 0.11.x and do a patch release of that, which will fix rust-elements 0.25+.

apoelstra (approved):
ACK 707e21404d677732e094cb1d8723cb77749345db; successfully ran local tests

PR #103, opened 2026-09-03
Bump secp256k1 dependency to 0.33

mllwchrry:
Bumps secp256k1 from 0.31.1 to ^0.33 and adapts to the 0.33 API:

- SecretKey: `as_ref()`/`from_byte_array` is now `as_secret_bytes()`/`from_secret_bytes`
- drop the removed `&Secp256k1` context arg from `from_secret_key`/`add_exp_tweak`
- replace deprecated context methods with free functions in tests/doctests (`generate_keypair`, `ecdsa::sign`, `ecdsa::verify`)
- stop forwarding the removed `secp256k1/hashes` feature.


src/zkp/ecdsa_adaptor.rs
```diff
@@ -179,7 +179,7 @@ impl EcdsaAdaptorSignature {
             ffi::secp256k1_ecdsa_adaptor_encrypt(
                 secp.ctx().as_ptr(),
                 &mut adaptor_sig,
-                sk.as_ref().as_ptr(),
+                sk.to_secret_bytes().as_ptr(),
```
apoelstra: In e44e5fcfd04d1c93da54389d7e28681b7015c7da:

This (and most other instances) are unsound. `to_secret_bytes` returns an array by value, which is dropped before the FFI function is called. The pointer returned by `as_ptr` is then dangling.

Please use `as_secret_bytes` instead of `to_secret_bytes`.

apoelstra: Several instances of this.

mllwchrry: Thanks, fixed in b2956a4.

apoelstra:
In b2956a4e8cd16f1cb2cbea92b9989d86f98bc535:

Please update the `Cargo-recent.lock` file as well.

Easiest way is to run

```
cp Cargo-recent.lock Cargo.lock
cargo check
cp Cargo.lock Cargo-recent.lock
```

Where `cargo check` can be pretty-much any cargo invocation. This will update the lockfile for the new rust-secp dependency without touching any other dependency verisons.

mllwchrry:
Updated `Cargo-recent.lock`. The only version changes are `secp256k1` and `secp256k1-sys`. The rest of the diff is just packages removed as a consequence of the bump.

apoelstra (approved):
ACK b2956a4e8cd16f1cb2cbea92b9989d86f98bc535; successfully ran local tests

apoelstra (approved):
ACK 6fa6e056a06aa60363222337d85636646f651d91; successfully ran local tests

PR #104, opened 2026-09-07
secp256k1-zkp-sys: multiple improvements

apoelstra:
This PR makes a pile of improvements to secp256k1-zkp-sys to give it a saner API.

It adds the crate to our workspace, which makes `cargo test` cover it; it fixes `cargo test` which was not compiling; it removes unused modules (all the ones whose logic we now reexport from rust-secp); it updates for some things that were deprecated by upstream; it cleans up the alloc/std feature-gating; it replaces some custom C-interop types with ones from `core::ffi` (which is stable for our MSRV, though not stable for rust-secp's MSRV).

I have a followup PR that will make more changes to secp256k1-zkp-sys, so we are not yet ready for release here.


secp256k1-zkp-sys/build.rs
```diff
@@ -33,26 +33,26 @@ fn main() {
...
         .define("ENABLE_MODULE_WHITELIST", Some("1"))
-        .define("ECMULT_GEN_PREC_BITS", Some("4"))
-        // TODO these three should be changed to use libgmp, at least until secp PR 290 is merged
-        .define("USE_NUM_NONE", Some("1"))
-        .define("USE_FIELD_INV_BUILTIN", Some("1"))
-        .define("USE_SCALAR_INV_BUILTIN", Some("1"));
+        // upstream sometimes introduces calls to printf, which we cannot compile
+        // with WASM due to its lack of libc. printf is never necessary and we can
+        // just #define it away.
+        .define("printf(...)", Some(""));
```
mllwchrry, line 46: This is right for WASM, but on fortified glibc it collides with the inline `printf` from the still present util.h:16 `#include <stdio.h>`, failing all optimized builds.

mllwchrry (commented):
The changes are good. But the CI is failing, and I found leftovers now that recovery has been removed:
- contrib/test.sh still lists recovery in FEATURES (fails the debug jobs).
- src/lib.rs still documents recovery in the crate feature list (cosmetic).

secp256k1-zkp-sys/build.rs
apoelstra: I'm just gonna drop this `printf` change. Unsure why it doesn't cause problems upstream.

